PR #2, opened 2026-07-15
feat: add DeepSeek provider support

Tian-Timm:
## Summary

- Add DeepSeek as a first-class provider using the existing OpenAI-compatible adapter
- Add `DEEPSEEK_API_KEY` configuration
- Use `deepseek-v4-pro` as the default main and small model
- Expose DeepSeek in Dashboard settings and cost estimates
- Add deterministic coverage for provider configuration

## Test plan

- [x] DeepSeek provider regression test passes
- [x] Deterministic test suite passes: 33 passed, 5 skipped
- [x] Ruff lint passes
- [x] No real API key is included

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 # ── Provider ────────────────────────────────────────────────────────────────
 # Pick ONE provider and paste its key below. That's the whole setup.
-#   anthropic (default) · openai · gemini · kimi · glm
+#   anthropic (default) · openai · gemini · deepseek · kimi · glm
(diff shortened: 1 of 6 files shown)


ShenSeanChen:
Merged — thanks again @Tian-Timm! Textbook provider addition: it reuses the existing OpenAI-compatible adapter, adds no new dependency, keeps the docs neutral, and ships a deterministic test for it. That's exactly the contribution pattern we're hoping for as Waku grows. Two thoughtful merges from you in a single day — thank you for helping build this.

PR #3, opened 2026-07-15
fix: read procedural skills as UTF-8

Tian-Timm:
Specify the Agent Skills text encoding so Windows installations with a GBK system locale can load the bundled UTF-8 SKILL.md files. Add a deterministic regression test that simulates a platform default unable to decode the file. Verified with the release gate: 33 passed, 5 skipped.

### DIFF
--- a/evals/deterministic/test_skill_encoding.py
+++ b/evals/deterministic/test_skill_encoding.py
@@ -1,0 +1,28 @@
+"""DETERMINISTIC EVAL — procedural skills use a portable text encoding."""
+
+from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


ShenSeanChen:
Merged — thank you @Tian-Timm! This is a real cross-platform bug: on a Windows box with a GBK system locale, the default encoding couldn't decode the UTF-8 SKILL.md files, so the agent couldn't load its own skills. Your one-line fix plus the monkeypatched regression test pin it exactly. This is Waku's very first community merge — genuinely appreciate you making it work for more people out of the box. Hope to see you around the repo.

PR #4, opened 2026-07-15
OpenRouter as a first-party provider, with an in-dashboard model picker

junlov:
## What

Adds `openrouter` to the provider table and lets the dashboard switch models without touching `.env`:

- **`loop/models.py`**: one new `PROVIDERS` entry (OpenAI wire format, `OPENROUTER_API_KEY`, free-tier default models). No new wire code; it rides the existing `OpenAICompatClient`.
- **`GET /api/models`**: the dashboard asks the active endpoint for its live catalog (stdlib `urllib`, 5-minute cache, brief failure caching so an unreachable catalog can't stall the poll). On OpenRouter each entry carries free / tool-support / reasoning / context / price metadata; native-wire providers fall back to their known defaults.
- **Settings tab, model catalog**: suggested picks per slot (transparent heuristics over tools/price/context/reasoning, labelled as heuristics, not a leaderboard), a substring filter with free-only and tool-capable-only toggles, and the full catalog grouped by vendor. Every row switches either slot in one click ("use" = loop model, "gate" = small model). The topbar model chip jumps here.
- **Settings form decluttered**: the manual model-id fields (type-to-search datalists) and the six BYOK key fields sit behind `<details>` disclosures, so the default view is provider + save; the keys disclosure auto-opens when the active provider has no key. The gate field explains itself inline (the small model that decides whether a message needs memory and distills chats into facts), and reasoning-model pills carry a tooltip saying why they misfit the gate slot.
- **Chat feedback while a model thinks**: "thinking..." ticks elapsed seconds, and past 20s explains that slow models queue and the turn errors at `WAKU_LLM_TIMEOUT` instead of hanging silently. Free-tier models make this state common enough to deserve honest UI.
- **Per-model cost**: the usage ledger already stored model ids, so cost estimates use the catalog's exact per-model pricing; `:free` ids read $0.
- **Eval fixes that fell out of this**: `HAS_KEY` and the release gate hardcoded `ANTHROPIC_API_KEY`, silently skipping live/judge evals for every other provider; they now key off the active provider. Un-skipping the live suite caught a real adapter bug (OpenRouter answers some rate limits with HTTP 200 + error body + `choices=null`, which became a bare `TypeError`), fixed to surface the endpoint's message. Plus 23 new offline provider-table evals pinning client wiring, key errors, pricing coverage, and the picker fallback for all six providers.

## Why

The five built-in providers all assume a paid key. OpenRouter's free tier makes the whole harness runnable for $0, which fits a teaching repo: clone, paste one free key, watch the loop. Its catalog is 340+ models, and Waku's two slots have opposite needs (the loop requires tool calling; the gate needs terse JSON on a 100-token budget, which reasoning models blow on chain-of-thought), so the picker encodes those constraints instead of letting users discover them the hard way.

## Testing

- `make gate`: **GATE OPEN**. 60 deterministic passed (including the live model-in-the-loop cases, on OpenRouter) and both judge evals passed, with the same endpoint serving as judge.
- `make lint` clean throughout; offline suite runs keyless and networkless (55 passed / 5 skipped).
- Live on the free tier: gate decisions, `save_note`, memory retrieval feeding `create_event` (the "Alex prefers mornings" demo books 9am), streaming dashboard chat, and paid-model traffic priced correctly in the ledger next to $0 free calls.
- Browser click-through of the final UI: catalog suggestions, vendor groups, both-slot switching, current/GATE pills, the collapsed disclosures, and the elapsed-seconds ticker during a live turn.
- Not tested live: real API traffic for the other five providers (no keys here). Their wire code is untouched except the no-choices guard, which turns a crash into a readable error. A smoke test with an Anthropic key before merging would be a sensible final check.

## Scope notes

Kept to the repo's bar: stdlib only, no new dependencies, no emojis. The eval/adapter fixes stand alone; happy to split them into a separate PR if you'd rather take those first.

Found while testing, may save others pain: reasoning-style models spend the gate's whole 100-token budget thinking out loud and never emit the JSON, so the catalog flags them and the gate suggestions exclude them; the gate default here is an instruct model (gemma).


junlov:
Pushed two commits since this went ready for review, both from a UX pass of actually using the page:

- `5e64a41` grouped catalog with per-slot suggestions (described in the body)
- `e84e634` settings/chat polish: themed filter controls, model-id + key fields behind disclosures (keys auto-open when the active provider has no key), inline explanation of the gate slot, reasoning-pill tooltip, an elapsed-seconds ticker on "thinking..." with a slow-model hint past 20s, and consistent chat-only wording

Body above is updated to match. Diff stands at 5 files, no new dependencies.

ShenSeanChen:
Merged — thank you @junlov, this is the biggest community contribution yet. The live model picker is exactly what people have been asking for (four of our first six PRs were providers — clearly everyone wants to choose their brain), and the engineering under it is genuinely careful: the 5-minute catalog cache with failure-caching so an unreachable endpoint can't stall the dashboard poll, per-model pricing instead of one number per provider, the free-tier $0 on-ramp defaults, generalizing live evals + the release gate to whatever provider is active, and the no-choices guard in the adapter. I brought your branch up to date with main (merge commit, your authorship preserved) — DeepSeek and MiniMax landed after you opened this, so the picker now fronts all 8 providers. It'll be on camera in the next video, credited. Hope to see you around the repo.

junlov:
Hey Shen, really enjoyed the way you built this repo. I'm current working on the scale branch on my fork where I'm turning this into a training course for myself similar to a KodeKloud or something. I'll be making a video series going through it and I'll share with you a bit of what I'm doing so you can see it and if you have any feedback, we can keep working on it.

Sent from [Proton Mail](https://proton.me/mail/home) for Android.

-------- Original Message --------

PR #5, opened 2026-07-16
fix: write procedural skills as UTF-8

Tian-Timm:
## Summary
- Write agent-created skills explicitly as UTF-8 before refreshing the skill loader.
- Preserve UTF-8 when installing community skills and saving dashboard edits.
- Add deterministic regression coverage for all three official `SKILL.md` write paths, independent of the host OS default encoding.

## Why

The loader now reads every `SKILL.md` as UTF-8, but the three official write paths still relied on the platform default encoding. On Windows with CP936/GBK, creating or saving a skill containing Chinese text or emoji could write a non-UTF-8 file and then fail when Waku immediately reloaded it.

## Test Coverage

- Agent `create_skill`: writes Chinese text and emoji, then refreshes and reloads the skill.
- `waku skill install`: mocks a UTF-8 download and verifies the installed file remains UTF-8.
- Dashboard `save_skill`: saves non-ASCII content and verifies the loader can read it.
- Tests force implicit `Path.write_text()` calls to fail, so Linux CI catches the Windows regression too.

## Pre-Landing Review

No issues found. No new dependencies, configuration, prompt behavior, or frontend behavior.

## Test plan

- [x] Confirmed all three new tests fail against the old implicit writes.
- [x] `python -m pytest -q evals/deterministic` — 37 passed, 5 skipped.
- [x] `python -m waku.ops.release_gate` — gate open; judge suite skipped without `ANTHROPIC_API_KEY`.
- [x] `python -m ruff check waku evals` — passed.
- [x] `python scripts/validate_skills.py` — passed.
- [x] Live CP936 reproduction: Chinese + emoji skill writes as UTF-8 and reloads successfully.

### DIFF
--- a/evals/deterministic/test_skill_encoding.py
+++ b/evals/deterministic/test_skill_encoding.py
@@ -1,8 +1,37 @@
 """DETERMINISTIC EVAL — procedural skills use a portable text encoding."""
 
+from io import BytesIO
 from pathlib import Path
+from types import SimpleNamespace
 
+import pytest
+
+from waku.memory.procedural import installer
 from waku.memory.procedural.loader import SkillLoader
+from waku.ops.dashboard import memory_action
+from waku.tools.memory_admin import make_create_skill_tool
+
+DESCRIPTION = "\u5904\u7406\u4e2d\u6587\u5468\u62a5"
+BODY = "\u7b2c\u4e00\u6b65\uff1a\u603b\u7ed3\u672c\u5468\u3002 \U0001f680"
+
+
+@pytest.fixture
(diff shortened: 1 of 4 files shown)


ShenSeanChen:
Merged — and that's three from you, @Tian-Timm. This one is the mark of a real contributor: you came back to finish your own fix, closing the write side (create_skill, the installer, the dashboard editor) so the whole SKILL.md round-trip is encoding-safe on any OS. The three-path regression test with Chinese text and emoji as the payload is exactly how we want this repo tested. Thank you for maintaining your corner of Waku.

PR #14, opened 2026-07-17
feat: add Discord gateway support: enable messaging Waku from Discord

### LINKED ISSUE #7: Discord gateway (one file, mirrors Telegram)
Add `waku/gateway/discord.py`: message-in → `waku.respond()` → message-out, using discord.py behind a `[discord]` extra.

**Reference to imitate:** [`waku/gateway/telegram.py`](https://github.com/ShenSeanChen/waku-agent/blob/main/waku/gateway/telegram.py) — same shape: own session_id (`discord`), env-gated token, daemon-thread `start_in_background()` optional.
**Verify:** chat with your bot; the Gateway tab tags messages `discord`.
**House rules:** one file, dep behind an extra, deterministic eval for anything testable offline, honest error strings. See CONTRIBUTING.md.

---

likweitan:
## Summary

Adds an optional Discord gateway so users can message Waku from Discord servers or direct messages.

## Changes

- Add waku.gateway.discord with message handling, typing indicators, reply delivery, and optional user allowlisting.
- Add the discord optional dependency backed by discord.py.
- Add `python -m waku discord` and make discord entry points.
- Start the Discord gateway alongside the dashboard when `DISCORD_BOT_TOKEN` is configured.
- Document `DISCORD_BOT_TOKEN`, `DISCORD_ALLOWED_USER`, and required Message Content Intent setup in `.env.example`.
- Isolate background Discord failures so they cannot stop the dashboard.

## Test plan

- [x] chat with your bot — passed.
- [x] Gateway tab message discord — passed.
image

- [x] Manually verified the committed implementation and configuration paths.
- [x] `python -m pytest -q evals/deterministic` — 88 passed.
- [x] `python -m ruff check waku evals` — passed.

Fixes #7


ShenSeanChen (changes requested):
tested with a real bot, every message crashes:

```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread.
```

`asyncio.to_thread` moves `respond()` onto a worker thread, but the sqlite connection is opened on the main thread in `_build_client`. telegram doesn't hit this because it calls `waku.respond()` directly.

simplest fix is to match telegram and drop the `to_thread`. the bot blocks during a turn, same as telegram does today.

if you'd rather keep it off the event loop, `db.connect()` takes `check_same_thread=False` (the dashboard uses it that way with a lock), but that needs plumbing through `Waku()` plus a lock, so i'd start simple.

everything else is right: mirrors telegram, dep behind the `[discord]` extra, and the message content intent note in .env.example is accurate, i hit exactly that.

likweitan:
Confirmed that `asyncio.to_thread` is the issue; dropping it and calling `respond()` directly (matching Telegram) fixes the issue.

Issue #9, opened 2026-07-17
Google Calendar adapter behind the same create_event schema


ShenSeanChen:
Swap the internals of [`waku/tools/calendar.py`](https://github.com/ShenSeanChen/waku-agent/blob/main/waku/tools/calendar.py) behind the SAME tool schema, env-gated (`WAKU_GOOGLE_CALENDAR=1`); the local .ics mock stays the default.

**Verify:** the deterministic eval still passes against the mock; live writes appear in Google Calendar.
**House rules:** same honest return strings, timeouts on every network call.

---
**Comment here before you start and I will assign it to you**, so two people never build the same thing (that happened on #12).

Tian-Timm:
Hi Sean, I’m interested in working on this. My proposed scope is to keep SQLite/ICS as the source of truth and add Google Calendar as an opt-in write target for create_event; list_events would remain local, and Google failures would return an honest partial-success message without rolling back the local event. Does that match your intent? Also, should attendee invitations be sent or suppressed?

ShenSeanChen:
assigned, and yes that matches. sqlite/ics stays the source of truth, gcal is an opt-in write target, list_events stays local.

the partial-success handling is exactly right, don't roll back the local event if google fails, just say so honestly in the return string.

on invites: suppress them by default (`sendUpdates=none`). the agent shouldn't email real people unless someone explicitly asks for it, and we can add a flag later. keep the dep behind a `[gcal]` extra and timeouts on every call, and the deterministic eval has to still pass with no network.

also reviewed your #16, it's good. needs a rebase, i restructured the frontend this week. details over there.

PR #16, opened 2026-07-18
fix: write JSONL traces as UTF-8

Tian-Timm:
Prevent Windows system encodings from breaking traces containing Chinese or emoji. Dashboard trace readers now use UTF-8, and legacy non-UTF-8 traces are reported without being modified. Verified with 117 deterministic tests, Ruff, the release gate, and a live CP936 round trip.

### DIFF
--- a/evals/deterministic/test_trace_encoding.py
+++ b/evals/deterministic/test_trace_encoding.py
@@ -1,0 +1,113 @@
+"""DETERMINISTIC EVAL — JSONL traces have one portable UTF-8 encoding."""
+
+from __future__ import annotations
+
+import json
+from datetime import datetime
+from pathlib import Path
+
+import pytest
+
+from waku.config import Settings
+from waku.ops.dashboard import collect, events_since
+from waku.ops.tracing import TraceEncodingError, Tracer
+
+
+MESSAGE = "处理中文日程 " + chr(0x1F680)
+
+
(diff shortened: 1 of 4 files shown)


ShenSeanChen:
reviewed this, 5/5 tests pass and the approach is better than what i asked for. refusing to append to a non-utf-8 file instead of guessing or rewriting it is the right call, and surfacing it in the dashboard rather than failing silently is the part i wouldn't have thought of.

one blocker: it conflicts now. i split `static/app.js` into `static/js/*.js` this week and moved things around in `dashboard.py`, so your `views.js` and `dashboard.py` hunks no longer apply. the logic should carry over unchanged, it's the surrounding lines that moved.

rebase on main and i'll merge. also assigned you #9 per your comment there.

Tian-Timm:
Rebased on the latest main, resolved the dashboard.py conflict while preserving the new frontend structure, and force-pushed the updated branch. The 5 targeted tests pass and Ruff is clean. Ready to merge.

ShenSeanChen (approved):
rebased cleanly and it's better than what i asked for. refusing to append to a non-utf-8 file instead of guessing, and surfacing the error in the dashboard rather than failing silently, is exactly right. 5/5 tests pass. thanks for the quick turnaround on the rebase.

PR #17, opened 2026-07-18
feat(cli): add /memory snapshot command

### LINKED ISSUE #11: /memory command in the CLI REPL
In the terminal REPL, `/memory` should show what Waku knows right now: semantic facts, recent episodes, and the unconsolidated-chat count. Pure SQLite reads — great for demos.

**Reference:** the REPL loop in [`waku/gateway/cli.py`](https://github.com/ShenSeanChen/waku-agent/blob/main/waku/gateway/cli.py) + queries in `waku/memory/`.
**Verify:** deterministic eval on a seeded temp home.

---

floze-the-genius:
## Summary

Closes #11.

Adds a read-only `/memory` command to the terminal REPL so users can inspect Waku's local state without invoking a model or mutating the database.

The bounded snapshot includes:

- the total semantic-fact count and 8 newest facts
- the total episode count and 5 newest episodes
- the number of unconsolidated chat messages

Stored text is rendered as literal `rich.Text`, so memory content cannot be interpreted as Rich markup.

## Testing

- `make gate`
  - 114 deterministic tests passed
  - 7 deterministic tests skipped
  - judge suite skipped because no Anthropic API key is configured
  - release gate open
- `.venv/bin/python -m ruff check waku/gateway/cli.py evals/deterministic/test_cli_memory.py`
- `git diff --check`

The deterministic eval seeds an isolated temporary Waku home and also covers an empty home. No network calls are made.


ShenSeanChen (approved):
tested against my real memory, works exactly as described. the unconsolidated count is a nice touch, it makes consolidation visible from the repl. merging.

PR #18, opened 2026-07-19
feat: add a configurable Notion episodic memory backend

### LINKED ISSUE #10: Notion episodic-memory adapter (same store interface)
Add `waku/memory/episodic/notion_store.py` implementing the same `add`/`search`/`recent` interface so episodes can live in a Notion database.

**Reference:** `SqliteEpisodeStore` in [`waku/memory/episodic/store.py`](https://github.com/ShenSeanChen/waku-agent/blob/main/waku/memory/episodic/store.py).
**Verify:** a deterministic eval against a stubbed Notion client (no network in CI).

---

wilenWang:
## Summary

Closes #10.

Issue #10 asks for `waku/memory/episodic/notion_store.py` implementing the
same `add` / `search` / `recent` contract as `SqliteEpisodeStore`, with a
deterministic test that does not call the Notion API.

This PR implements that adapter and adds the integration needed to make it a
usable Waku feature. A store class on its own would otherwise remain unused by
`Memory`, require manual configuration, and leave the dashboard unable to show
or manage Notion-backed episodes.

### Notion episodic store

- Added `NotionEpisodeStore` with compatible `add`, `search`, and `recent`
  methods.
- Stores each episode as a Notion page with `Name` (title) and `Summary`
  (rich text) properties.
- Resolves the selected database to its data source and uses the current
  Notion data-sources API for writes and paginated reads.
- Adds `list` and `delete` support so existing Waku memory operations can
  inspect and archive Notion episodes.
- Uses local keyword matching and ISO-8601 timestamp ordering to preserve the
  expected episodic-memory behavior without making retrieval non-deterministic.
- Accepts either a raw database ID or the database URL copied from Notion; the
  URL path is normalized to the database ID and its `v=` view ID is ignored.

### Runtime and configuration

- Added the optional `notion-client` dependency via `waku-agent[notion]`.
  Default installations remain dependency-free.
- Added `WAKU_EPISODIC_STORE=sqlite|notion`, with SQLite remaining the default.
- Updated `Memory` to instantiate the configured episodic store.
- Added Dashboard Settings controls for backend selection, integration token,
  and a pasteable Notion database link.
- Persists configuration locally and rebuilds the in-process agent after a
  successful switch.

### Dashboard and memory operations

- Shows whether episodic memory is sourced from SQLite or Notion.
- Lists Notion episodes in the existing memory view.
- Archives the Notion page when an episode is deleted from the dashboard or
  memory-management tool.
- Keeps the rest of the dashboard available if a Notion read fails, surfacing
  the backend error in the episodic-memory view instead.

### Tests and documentation

- Added deterministic tests with a stubbed `notion_client.Client`, so CI makes
  no Notion network requests.
- Covers creation, query ordering, keyword search, missing configuration,
  lazy imports, pagination-compatible queries, backend selection, dashboard
  collection, and string Notion page IDs.
- Documented installation and environment configuration in `.env.example`.

## Why This Goes Beyond the Minimal Issue Text

The adapter is the core deliverable for #10, but it is not enough for users to
run the feature end to end. The additional changes make the backend selectable,
configurable, observable, and manageable through the existing product flows:

1. `Memory` must select the adapter instead of always constructing SQLite.
2. Users need a supported way to configure the backend, token, and database.
3. Existing dashboard and admin flows need to handle Notion's string page IDs.
4. A Notion outage must not take down unrelated dashboard data.
5. Requiring users to manually extract an ID from a copied Notion URL creates
   unnecessary setup friction.

## How To Use

```bash
pip install -e '.[notion]'
```

Create a Notion database with the following properties, then share the original
database with the integration:

- `Name`: title
- `Summary`: rich text

Create a Internal Connection on Notion to get the `Notion_Token`:
follow [this](https://developers.notion.com/guides/get-started/internal-connections)

Use copy link to Get `NOTION_EPISODES_DATABASE_ID`
image

Configure it in Dashboard Settings, 
image

or set:

```bash
WAKU_EPISODIC_STORE=notion
NOTION_TOKEN=...
NOTION_EPISODES_DATABASE_ID='https://app.notion.com/p/<copied-database-link>'
```

The setting also accepts a raw database ID for existing configurations.

## Verification

```bash
pytest evals/deterministic/test_notion_episodes.py \
       evals/deterministic/test_episodic_store_switch.py -q

pytest evals/deterministic/test_static_assets.py -q
```

The targeted Notion/backend test suite passes without network access.


wilenWang:
@ShenSeanChen This PR is larger than the previous ones and introduces quite a few more changes. I understand these changes are needed to make the feature work end-to-end. There is no rush to merge this PR, so please take your time to review and discuss any concerns or suggestions.

ShenSeanChen (approved):
thanks for the note about taking my time, i did, and this deserved it. tested end to end against a real notion database, not just the fixtures: chatted through a consolidation cycle and watched the episode land in notion, then watched the dashboard read it back. 22/22 tests pass and it merges clean.

the giveaway that it's really switching: my sqlite has 11 episodes and the dashboard showed 1 after the switch. it's reading notion, not the local db.

what i checked specifically, since this is the largest pr the repo has taken:

- `notion-client` sits behind a `[notion]` extra, so the core stays stdlib + anthropic/openai. that's the rule i care most about
- `episodic_store` mirrors the existing `semantic_store` setting exactly, so there's one pattern to learn rather than two
- `normalize_database_id` accepting a pasted url is a small kindness that saved me real confusion
- the failure isolation is the best part. i hit both failure modes for real (extra not installed, then database not shared with the integration) and neither took the dashboard down — `episodes_payload` returned the error as data and the ui showed it. that's how it should work
- you caught the id type change too: quoting `delMem('delete_episode','${e.id}')` and widening the tool schema to `["integer","string"]`. sqlite ids are ints, notion page ids are uuid strings. easy to miss

one thing worth a follow-up, not a blocker: `NotionEpisodeStore()` is constructed inside `collect()`, which runs on every dashboard auto-refresh. with the notion backend on that's a new client plus a network round-trip every few seconds just from having the tab open. worth caching the client or the result. happy to file it separately.

merging. this is the model for how to add a backend to waku — mind if i point future contributors at it as the reference?

wilenWang:
> thanks for the note about taking my time, i did, and this deserved it. tested end to end against a real notion database, not just the fixtures: chatted through a consolidation cycle and watched the episode land in notion, then watched the dashboard read it back. 22/22 tests pass and it merges clean.
> 
> the giveaway that it's really switching: my sqlite has 11 episodes and the dashboard showed 1 after the switch. it's reading notion, not the local db.
> 
> what i checked specifically, since this is the largest pr the repo has taken:
> 
> * `notion-client` sits behind a `[notion]` extra, so the core stays stdlib + anthropic/openai. that's the rule i care most about
> * `episodic_store` mirrors the existing `semantic_store` setting exactly, so there's one pattern to learn rather than two
> * `normalize_database_id` accepting a pasted url is a small kindness that saved me real confusion
> * the failure isolation is the best part. i hit both failure modes for real (extra not installed, then database not shared with the integration) and neither took the dashboard down — `episodes_payload` returned the error as data and the ui showed it. that's how it should work
> * you caught the id type change too: quoting `delMem('delete_episode','${e.id}')` and widening the tool schema to `["integer","string"]`. sqlite ids are ints, notion page ids are uuid strings. easy to miss
> 
> one thing worth a follow-up, not a blocker: `NotionEpisodeStore()` is constructed inside `collect()`, which runs on every dashboard auto-refresh. with the notion backend on that's a new client plus a network round-trip every few seconds just from having the tab open. worth caching the client or the result. happy to file it separately.
> 
> merging. this is the model for how to add a backend to waku — mind if i point future contributors at it as the reference?

Absolutely — I’d be honored for future contributors to use it as a reference.

Thank you for taking the time to test it so thoroughly against a real Notion database, especially the failure paths and backend switching. Your feedback is incredibly helpful, and I’m glad the implementation held up end to end.

The repeated client construction and refresh-time network call is a good catch. I agree it’s better handled as a separate follow-up, and I’d be happy to work on it or help review the issue.

Thanks again for the thoughtful review and for merging it.

PR #13, opened 2026-07-17
feat: add trace pretty-printer

c436zhan:
## Summary
Adds `python -m waku.ops.show_trace [file]` to render Waku JSONL traces as a readable indented terminal timeline.

## Root cause
Waku already writes useful chronological trace events, but there was no terminal-friendly way to inspect those JSONL files without reading raw records.

## Changes
- Add `waku.ops.show_trace` with JSONL parsing, event-specific formatting, and a small CLI wrapper using the existing `rich` dependency.
- Add deterministic eval coverage for the formatter and JSONL reader.

## Testing
- `.venv/bin/python -m pytest -q evals/deterministic/test_show_trace.py`
- `make eval`
- `make lint`
- `.venv/bin/python -m waku.ops.show_trace /tmp/waku-show-trace-smoke.jsonl`

Fixes #12

PR #21, opened 2026-07-22
Show several daily traces at once

### LINKED ISSUE #19: show_trace: --days N to render several daily traces at once
`python -m waku.ops.show_trace` (added in #15) renders one daily trace file. Reading a week means a shell loop:

```bash
for f in .waku/traces/*.jsonl; do python -m waku.ops.show_trace "$f"; done
```

Add `--days N` (and/or accept multiple paths) so a stretch of days renders as one timeline, with a clear separator per day.

**Verify:** a deterministic eval in `evals/deterministic/` covering multiple files and a day with no trace.
**House rules:** stdlib only, stream rather than loading files into memory (same as #15).

Follow-up to #12 / #15.

---

c436zhan:
## Summary
Add multi-day trace rendering to `python -m waku.ops.show_trace` so a recent window of daily JSONL traces can be read in one pass.

## Root cause
`show_trace` only accepted zero or one trace path, so reading several daily trace files required a shell loop and skipped over missing days instead of making the gap visible.

## Changes
- Add `--days N` to render the last N dated trace files from `.waku/traces` with a separator for each day.
- Accept multiple explicit trace paths and render each with a visible separator.
- Add deterministic coverage for multiple days including a missing trace day.

## Testing
- `.venv/bin/python -m pytest -q evals/deterministic/test_show_trace.py`
- `.venv/bin/python -m pytest -q evals/deterministic`
- `.venv/bin/python -m ruff check waku/ops/show_trace.py evals/deterministic/test_show_trace.py`
- `.venv/bin/python -m ruff format --check waku/ops/show_trace.py evals/deterministic/test_show_trace.py`
- `make gate`
- `make lint`

Fixes #19


ShenSeanChen (changes requested):
this works and i'm glad you took it on — `--days N` renders the range, multiple explicit files work, and `--days 0` is rejected cleanly. tests pass.

one change before merging. `--days` walks calendar days rather than the files that exist, so on any real machine it prints a "No trace file found" line for every day the agent wasn't run. i have 5 trace files; `--days 30` printed 25 of those error lines and buried the 5 real days in the middle. nobody runs the agent every day, so this hits everyone.

could you skip missing days silently when `--days` is used? optionally a single summary line at the end like "20 of 30 days had no trace" is nice, but even just skipping them is enough. keep the "No trace file found" message for the case where someone names a file explicitly that doesn't exist — that one's still useful.

everything else is good. thanks again for picking this up after #13.

c436zhan:
Thanks for the review. Addressed now in [45a5f13](https://github.com/ShenSeanChen/waku-agent/pull/21/commits/45a5f13cb9777ad7b249ecad7cf3f7b1e2fc0347)

ShenSeanChen:
thanks for turning that around, and sorry for the wait. CI was sitting unapproved on my side, which is why you saw no checks. it's running now.

re-reviewing today.

ShenSeanChen (approved):
tested against my real traces: `--days 30` over 7 trace files now prints 7 day headers and zero "No trace file found" lines. that was the whole ask and it's exactly right. 4/4 tests pass.

thanks for sticking with this one through the review round and the CI wait.

PR #22, opened 2026-07-22
fix: cache the Notion episodic client across dashboard refreshes

### LINKED ISSUE #20: Notion episodic backend rebuilds its client on every dashboard refresh
With `WAKU_EPISODIC_STORE=notion`, `episodes_payload()` in `waku/ops/dashboard.py` constructs a fresh `NotionEpisodeStore()` and queries Notion on every `collect()` — which runs on each dashboard auto-refresh. Leaving the tab open means a network round-trip to Notion every few seconds, which will hit rate limits and adds latency to every refresh.

Cache the client (and probably the result, with a short TTL) instead of rebuilding per refresh. The sqlite path is a local query so it doesn't have this problem.

**Verify:** a deterministic eval asserting the Notion client is constructed once across repeated `collect()` calls (mock the client — no network in tests).
**House rules:** a Notion outage must still degrade gracefully rather than take the dashboard down, as it does today.

Follow-up to #18.

wilenWang:
Closes #20

## Background

This is a follow-up to #10 and the corresponding PR #18, which introduced a configurable Notion episodic memory backend.

After #18 was merged, [this review comment](https://github.com/ShenSeanChen/waku-agent/issues/20) pointed out that with **WAKU_EPISODIC_STORE=notion**, **collect()** constructs a fresh **NotionEpisodeStore** on every dashboard auto-refresh. Because the dashboard polls roughly every 5 seconds, merely leaving the tab open creates a new Notion client plus a network round-trip several times a second. This is both slow and a quick way to hit Notion rate limits.

## What changed

- **Process-wide client cache** (waku/ops/dashboard.py)
  - Added **_get_notion_store()** so the **NotionEpisodeStore** instance is built once per process. Its constructor itself calls Notion (data-source resolution), so repeated construction was the main cost.

- **30-second TTL for episode query results** (waku/ops/dashboard.py)
  - Added **_notion_episodes** with a short TTL. The dashboard polls ~every 5s; caching the list for 30s cuts Notion queries by roughly 6× while keeping the data fresh enough for display.

- **Reuse the cached store in tools_info()** (waku/memory/__init__.py)
  - Added an optional **episode_store** parameter to **Memory**. **tools_info()** now passes in the dashboard's cached store instead of letting **Memory()** build a second fresh client per poll.

- **Cache invalidation on delete** (waku/ops/dashboard.py)
  - **memory_action(delete_episode)** clears **_notion_episodes** so deletions appear immediately instead of lingering for up to 30s.

- **Graceful degradation during a Notion outage**
  - If the TTL has expired and Notion is unreachable, **collect()** returns the last cached episode list with an error field set, rather than blanking the tab.

## Tests

Added four deterministic evals in evals/deterministic/test_episodic_store_switch.py:

1. **test_collect_builds_notion_client_once_across_refreshes** — confirms one client construction and one query across three **collect()** calls.
2. **test_collect_refetches_after_ttl_but_reuses_client** — confirms TTL expiry triggers a refetch while still reusing the cached client.
3. **test_collect_serves_stale_episodes_during_notion_outage** — confirms cached data is served when Notion is down.
4. **test_delete_episode_busts_the_episodes_cache** — confirms deletion invalidates the cache.

All tests use a mocked Notion client, so they run without network access.

## Verification

- [x] Dashboard auto-refresh with **WAKU_EPISODIC_STORE=notion** no longer shows repeated Notion client constructions.
- [x] Episode list updates within 30s and reflects deletions immediately.
- [x] Tab remains populated during a simulated Notion outage.

### DIFF
--- a/evals/deterministic/test_episodic_store_switch.py
+++ b/evals/deterministic/test_episodic_store_switch.py
@@ -19,8 +19,12 @@ class _FakeNotionClient:
     same rows as one created in the test body."""
 
     _pages: list[dict] = []
+    _init_count = 0     # Client constructions — the dashboard must build ONE
+    _query_count = 0    # data-source queries — the result cache throttles these
+    _query_fails = False   # simulate a Notion outage after a successful fetch
 
     def __init__(self, auth: str | None = None) -> None:
+        type(self)._init_count += 1
         self.auth = auth
         self.databases = types.SimpleNamespace(retrieve=self._retrieve)
         self.data_sources = types.SimpleNamespace(query=self._query)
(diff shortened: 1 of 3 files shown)


ShenSeanChen (approved):
tested, 15/15 pass and it's exactly the fix for #20 — the notion client is built once and reused instead of reconstructed on every dashboard refresh. clean and targeted. thanks for turning this around so fast.

wilenWang:
🚀🚀🚀！

PR #25, opened 2026-07-22
fix: explicit UTF-8 encoding for dashboard read_text() calls

### LINKED ISSUE #23: [Bug] Windows UTF-8 Encoding Degradation
All `read_text()` calls silently degrade in Windows to cp1252, which cannot handle emojis. This breaks the program on startup in Windows. Not sure if you intend to add Windows support but the app looks great and I'd love to help debug Windows behavior when I can.

luminary19:
Added explicit UTF-8 typing for `read_text()` calls. This fixes the `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 2393: character maps to <undefined>` runtime error.

Fixes #23


ShenSeanChen (changes requested):
thanks for this, and for filing #23 alongside it. hardening every `read_text()` in the dashboard against non-ascii bytes on windows is a real fix and most of these 11 spots weren't covered anywhere else — usage.jsonl, eval_report.json, SOUL.md, mcp.json, models.json, the outbox. good catch.

two things before it can merge:

1. it conflicts now. #16 landed a few hours ago and rewrote the same trace-read line in `collect()` that you also touch. please rebase on latest main and drop just that one hunk (the `for line in path.read_text()` inside the `trace_files` loop) — #16 owns that line now. your other 10 hunks don't overlap and should apply cleanly.

2. worth knowing for the trace/eval loops specifically: `read_text(encoding="utf-8")` still raises `UnicodeDecodeError` and takes the whole dashboard down if a file has a bad byte. #16 handles that case by catching it, showing the error in the ui, and refusing to guess — see `iter_trace_lines` in `waku/ops/tracing.py`. not asking you to redo everything that way, the encoding fix alone is a real improvement. but for the files that could realistically be corrupt (traces, ledgers), wrapping the read the way #16 does is the more robust pattern.

rebase + drop the one hunk and i'll merge. this closes #23.

luminary19:
Yep done @ShenSeanChen .

ShenSeanChen:
thanks, and sorry for the slow turnaround. CI was blocked on my approval this whole time, that's on me. it's running now.

reviewing today. @drakegriffith's #28 covers the same fix everywhere outside dashboard.py, so yours lands first and his goes on top.

ShenSeanChen (approved):
rebased and the overlapping hunk is gone. merging. thanks for the patience on this one, the CI silence was my fault not yours.

PR #26, opened 2026-07-22
feat: add WhatsApp gateway via Meta Cloud API webhook

### LINKED ISSUE #8: WhatsApp gateway (Meta Cloud API, webhook-based)
Add `waku/gateway/whatsapp.py` on the Meta Cloud API (webhook-in → `waku.respond()` → send-out). Document the Meta setup pain honestly in the module docstring — that honesty IS the contribution.

**Reference:** [`waku/gateway/telegram.py`](https://github.com/ShenSeanChen/waku-agent/blob/main/waku/gateway/telegram.py).
**Verify:** message your own WhatsApp number; Gateway tab tags `whatsapp`.
**House rules:** one file, deps behind a `[whatsapp]` extra, no secrets in code.

---

epicnellson:
Fixes #8

Summary

Adds a WhatsApp gateway to waku-agent using the Meta Cloud API, mirroring the existing Telegram gateway architecture exactly. Incoming messages arrive via webhook, get passed to waku.respond(), and replies are sent back via the Meta messages API.

Changes

waku/gateway/whatsapp.py (new file — 243 lines)

GET /webhook handles Meta's one-time verification challenge via VERIFY_TOKEN
POST /webhook parses incoming messages, calls waku.respond(), sends reply via _send_message()
main() — standalone waku whatsapp command on port 5000
start_in_background() — daemon thread for dashboard integration
80+ line docstring documenting every Meta gotcha honestly: token expiry, sandbox limits, business verification, ngrok URL churn, 24h message window, rate limits

pyproject.toml

Added [whatsapp] optional extra with httpx>=0.25

Makefile

Added make whatsapp target and .PHONY entry

.env.example

Added WHATSAPP_TOKEN, WHATSAPP_PHONE_NUMBER_ID, WHATSAPP_VERIFY_TOKEN, WHATSAPP_ALLOWED_PHONE with explanatory comments

waku/__main__.py

Added waku whatsapp route and docstring entry

waku/config.py

Added whatsapp_token and whatsapp_phone_number_id to Settings dataclass

waku/ops/dashboard.py

WhatsApp background thread starts alongside Telegram when waku dashboard runs
Usage
bash
pip install -e '.[whatsapp]'
# Add to .env:
# WHATSAPP_TOKEN=your_token
# WHATSAPP_PHONE_NUMBER_ID=your_id
# WHATSAPP_VERIFY_TOKEN=any_secret_string
# WHATSAPP_ALLOWED_PHONE=your_whatsapp_number
make whatsapp
Local testing with ngrok
bash
ngrok http 5000
# Set the ngrok URL as webhook in Meta developer portal
Lint

Ruff: All checks passed ✅

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -160,6 +160,16 @@ OPENCODE_GO_API_KEY=
 # DISCORD_REQUIRE_MENTION=1      # in a channel, only reply when @-mentioned
 # DISCORD_MAX_TURNS_PER_HOUR=30  # hard spend ceiling across all users
 # DISCORD_HOME=.waku-discord     # a SEPARATE memory — use this for any shared server
+# ── WhatsApp gateway (optional): pip install -e '.[whatsapp]' ──────────────
+# Meta Cloud API access token (temporary: 24h, permanent via System User).
+# See waku/gateway/whatsapp.py docstring for full setup guide.
+# WHATSAPP_TOKEN=
+# Phone Number ID from WhatsApp > Getting Started in your Meta app dashboard.
+# WHATSAPP_PHONE_NUMBER_ID=
+# Any random string — you'll enter it in Meta's webhook config page.
+# WHATSAPP_VERIFY_TOKEN=
+# Optional: lock the gateway to one phone number (without '+' prefix).
+# WHATSAPP_ALLOWED_PHONE=
 
 # ── Supabase vector memory (optional): pip install -e '.[supabase]' ────────
 # The launch-rag upgrade path: run sql/init_supabase.sql on a fresh project.
--- a/.env.example
+++ b/.env.example
@@ -166,6 +166,10 @@ OPENCODE_GO_API_KEY=
 # WHATSAPP_TOKEN=
 # Phone Number ID from WhatsApp > Getting Started in your Meta app dashboard.
 # WHATSAPP_PHONE_NUMBER_ID=
+# App Secret from your app's Settings → Basic → App Secret.
+# Required: Meta signs every webhook with this; without it, anyone who finds
+# your URL can forge payloads and drive the agent.
+# WHATSAPP_APP_SECRET=
 # Any random string — you'll enter it in Meta's webhook config page.
 # WHATSAPP_VERIFY_TOKEN=
 # Optional: lock the gateway to one phone number (without '+' prefix).
--- a/.venv
+++ b/.venv
@@ -1,0 +1,1 @@
+/Users/shenseanchen/Developer/waku-agent/.venv
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@
 PY := $(shell [ -x .venv/bin/python ] && echo .venv/bin/python || echo python)
 
 .PHONY: run voice telegram discord brief dashboard trace eval eval-judge gate lint
+.PHONY: run voice telegram whatsapp brief dashboard trace eval eval-judge gate lint
 
 run:            ## chat with Waku in the terminal
 	$(PY) -m waku
@@ -21,6 +22,8 @@ telegram:       ## phone → laptop (needs TELEGRAM_BOT_TOKEN in .env)
 
 discord:        ## Discord → laptop (needs DISCORD_BOT_TOKEN in .env)
 	$(PY) -m waku discord
+whatsapp:       ## WhatsApp → laptop (needs WHATSAPP_TOKEN in .env, public URL)
+	$(PY) -m waku whatsapp
 
 brief:          ## morning briefing from calendar + mail + memory
 	$(PY) -m waku brief
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ gcal = [
 # Optional gateway: message Waku from Discord
 discord = [
     "discord.py>=2.4",
+]
 # Optional gateway: WhatsApp via Meta Cloud API (needs a public URL — see docstring)
 whatsapp = [
     "httpx>=0.25",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,6 @@ gcal = [
     "google-api-python-client>=2.0",
     "google-auth>=2.0",
     "google-auth-httplib2>=0.2",
-    "google-auth-oauthlib>=1.0",
     "httplib2>=0.22",
     # The browser consent flow. Without it the only way in is `gcloud auth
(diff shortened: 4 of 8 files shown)


ShenSeanChen:
sorry for the long silence on this, and for two of the CI failures being mine.

the first one was my fault twice over: CI was never approved on my side (fixed, it runs now), and when it did run it failed on `# noqa: BLE001` in your gateway. that directive was correct when you wrote it; i later added BLE001 to the global ignore list, which turned every such comment into an unused-noqa error in a file you didn't touch. i've turned RUF100 off for that reason and pushed it, so that part is gone.

what's left is real and small, one rule:

```
I001 Import block is un-sorted or un-formatted
```

from the repo root:

```
ruff check waku --fix
```

then push. that should be the last thing between this and a merge. #14 (Discord) just went green through the same path, so the shape is proven.

ShenSeanChen (changes requested):
reviewed properly this time. the shape is right: dep behind a `[whatsapp]` extra, empty defaults, `hub.verify_token` checked with a 403, an allowed-sender check mirroring telegram, and a fast 200 so Meta doesn't disable the webhook. that last detail says you've actually run this.

one blocking change, and it's a security one.

**the POST endpoint has no signature verification.** telegram and discord dial outward, so nothing can reach them. a whatsapp webhook has to be publicly reachable — which means anyone who finds the URL can POST a forged payload and drive the agent: run tools, write to the calendar, spend tokens. the allowed-sender check doesn't stop it, because `msg["from"]` comes out of the same forged body.

Meta signs every webhook with `X-Hub-Signature-256: sha256=<hmac>`, keyed on the app secret. verifying it is the control that makes this safe:

```python
import hashlib, hmac

app_secret = os.getenv("WHATSAPP_APP_SECRET", "")
sig = self.headers.get("X-Hub-Signature-256", "")
expected = "sha256=" + hmac.new(app_secret.encode(), body, hashlib.sha256).hexdigest()
if not hmac.compare_digest(sig, expected):
    self._set_json(403); self.wfile.write(b"{}"); return
```

`hmac.compare_digest`, not `==` — timing. and please make it fail closed: if `WHATSAPP_APP_SECRET` isn't set, refuse to start, the same way you already refuse without a verify token. a webhook that silently accepts unsigned posts is worse than one that won't boot.

also one lint failure, unrelated:

```
I001 waku/gateway/whatsapp.py:127 — Import block is un-sorted
```
`ruff check waku --fix` from the repo root.

worth adding a deterministic eval too: a forged POST with a bad signature gets 403 and never reaches `respond()`. that's the regression case for exactly this.

two of your earlier CI failures were mine, not yours — CI sat unapproved for days, and my own ignore-list change turned your `# noqa: BLE001` into an error. both fixed. sorry for the wait.

ShenSeanChen:
the signature work is right, and i tested it properly this time. forged post, no signature, and a valid hmac made with the wrong secret all get 403 and never reach `respond()`. `compare_digest` instead of `==`, and it refuses to boot without `WHATSAPP_APP_SECRET`. that is exactly what i asked for, thank you.

but i ran it end to end with a real agent and hit a crash on every message that passes the guard:

```
sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in
that same thread. The object was created in thread id 8397790848 and this is
thread id 6116159488.
  File "waku/gateway/whatsapp.py", line 213, in do_POST
    result = waku.respond(
```

`ThreadingHTTPServer` (line 261) handles every request on a new thread, but `Waku()` is built once on a different one, and its sqlite connection goes with it. so the webhook 200s and the message is silently dropped. hits both `main()` and `start_in_background()`.

we already solved this for the dashboard in `waku/ops/browser_agent.py` — shareable connection plus a lock so turns serialise:

```python
from waku.config import load_settings
from waku.db import connect

s = load_settings(); s.ensure_home()
waku = Waku(settings=s, conn=connect(s.home, check_same_thread=False))
waku_lock = threading.Lock()

# then in do_POST:
with waku_lock:
    result = waku.respond(text, observer=_observer, source="whatsapp")
```

i applied exactly that locally and the whole thing works. two messages in, real gate decisions, a real tool call, and memory recall on the second one:

```
--> remember that raj plays tennis on saturdays
    gate . skip - User storing new memory
    tool . save_note({'subject': 'raj', ...}) -> Saved to memory under 'raj'
<-- Got it, i'll remember that Raj plays tennis on Saturdays.

--> what do you know about raj?
    gate . retrieve - User asking about specific person
<-- Just one thing so far: Raj plays tennis on Saturdays.
```

two other things, neither urgent:

- your branch is about 30 commits behind main now, and main requires branches to be up to date before merging. so a rebase before the next push. sorry, that is on me for moving fast this week.
- the 2 ruff failures in CI are in `scripts/demo_seed.py`, not your file. i fixed those on main, so the rebase clears them. `whatsapp.py` itself is ruff clean.

fix the thread thing and rebase and i think this is ready. the deterministic eval for the forged-post case i can add myself after, do not let that hold you up.

epicnellson:
Thread safety fixed and rebased onto latest main.

ShenSeanChen:
i rebased this for you rather than asking again. the conflicts were mine, not yours — main moved 58 commits under this PR while it was open (the dashboard got split into six modules, a discord gateway landed, google calendar work landed), and asking you to rebase a third time for churn i caused was not a reasonable thing to do.

your three commits are untouched and still yours. the fourth commit is only the conflict resolution.

rather than merging 58 commits of stale history i cherry-picked your three onto current main. every conflict was the same shape: main added discord where you added whatsapp. they are siblings, not alternatives, so both sides stayed everywhere — makefile targets, cli dispatch, .env.example, the extras, dashboard startup.

three things the rebase surfaced, all fixed in that commit and none of them yours:

1. `pyproject.toml` listed `google-auth-oauthlib` twice. PR #38 and my own gcal work each added it and neither of us noticed.
2. the `discord` extra lost its closing bracket where a conflict boundary cut through the array, which made the whole file unparseable as TOML.
3. keeping both sides fused the discord and whatsapp startup blocks into one `try` with two `except`s. each gateway has its own `try` now, because a discord failure must not silently skip whatsapp.

and i re-ran the attack after the rebase rather than assuming it survived. no signature, forged signature, and a valid hmac made with the wrong secret all get 403 with zero agent invocations. only a correctly signed payload reaches `respond()`. 291/291 offline deterministic, ruff clean.

thank you for sticking with this one. it sat for nine days, you fixed both things i asked for, and the delay was mine every time.

PR #27, opened 2026-07-23
feat: add opt-in Google Calendar event sync

### LINKED ISSUE #9: Google Calendar adapter behind the same create_event schema

Tian-Timm:
Closes #9

## Summary
- keep SQLite and calendar.ics as the source of truth while optionally mirroring create_event writes to Google Calendar
- preserve the existing tool schema and local list_events behavior
- use ADC, a 30-second authorized HTTP transport, static discovery, and sendUpdates=none
- report Google failures honestly without rolling back the local event
- isolate deterministic evals and dashboard compare runs from real Google Calendar writes
- document the [gcal] extra and setup flow

## Verification
- `15 passed, 11 skipped` — Google/Apple/tool-trigger deterministic tests
- `193 passed, 11 skipped, 3 deselected` — full deterministic suite on Windows, excluding three existing tests that hard-code the unavailable `python3` executable
- `ruff check waku evals`
- installed and imported google-api-python-client 2.198.0, google-auth 2.56.2, google-auth-httplib2 0.4.0, and httplib2 0.32.0
- built the real Calendar v3 static client request as `POST /calendar/v3/calendars/primary/events?sendUpdates=none` without discovery network access

## Live verification
- Not run: this machine does not have project-specific Google Calendar ADC configured, so no real calendar was mutated during development.


ShenSeanChen (approved):
merging. this hits every point from the #9 scope discussion and then some:

- `google-api-python-client` behind a `[gcal]` extra, core stays stdlib + the two API clients
- opt-in: `Settings().google_calendar is False` by default, and pinned by a test
- sqlite/ics stays the source of truth; a Google failure does NOT roll back the local event, and the return string says "still in the local calendar" instead of pretending
- `sendUpdates=none`, so the agent never emails real attendees on its own
- 7/7 tests pass

the detail i want to call out: you pass `google_calendar=False` explicitly in BOTH arena `Settings(...)` constructions. those two lines are the arena's per-race isolation guarantee, and i've already been bitten once by a global switch leaking into races. adding your flag there rather than letting it inherit is exactly right, and it's the kind of thing that's invisible unless you've read that code.

thanks, and sorry for the wait.

PR #28, opened 2026-07-23
fix: explicit UTF-8 encoding for all text file I/O outside the dashboard

drakegriffith:
Addresses the rest of #23 (companion to #25, which covers the dashboard). Python's text-mode file APIs default to the locale encoding — cp1252 on Windows — so memory, soul, calendar, trace, and eval files all degrade on non-ASCII content. Same fix already merged for traces in #16.

21 call sites, 13 files, no behavior change on macOS/Linux. `make eval`: 189 passed. Deliberately no dashboard.py changes to avoid conflicting with #25.


ShenSeanChen (approved):
merging. this is the sweep i wouldn't have found by eye: 21 sites across 13 files, every one a mechanical `encoding="utf-8"`, nothing else touched. and thanks for deliberately staying out of dashboard.py so @luminary19's #25 could land first, that sequencing worked exactly as planned.

#25 is in, so this goes on top. closes #23.

PR #29, opened 2026-07-23
feat: add meeting-prep community skill (first skills/community entry)

drakegriffith:
Taking you up on the standing invitation in the README — this would be the first skill in `skills/community/`.

**meeting-prep**: "prep me for my meeting with Alex" → finds the event via `list_events`, pulls what memory knows about the attendees, optionally one `search_web` if they're public-facing, and writes a skimmable prep card (offered to `save_note`). It's the natural third member of the schedule-meeting / weekly-brief family, and it leans on the memory pillar — the prep is only as good as what Waku remembers, which makes it a nice demo of why the memory layer exists.

Core tools only, so it runs on a default install (no Apple/env-gated tools). `scripts/validate_skills.py` passes (3 skills found); body is 39 lines, under the ~60-line guidance. `make eval`: all deterministic evals pass.


ShenSeanChen (approved):
merged. this is a model skill contribution: it only calls tools that exist (list_events, save_note, search_web), the edge-case table covers the cases i'd have found the hard way, and "ask how the Berlin launch went beats discuss recent developments" is the line that makes the whole thing work.

note for anyone reading this later: CI failed at first with 95 ruff errors in files this PR never touches. that was a stale base, not the contribution. updating the branch made it green.

PR #36, opened 2026-07-27
Refresh Waku architecture whiteboard

AllenHK:
## Summary\n\n- replace the superseded Hermes-era architecture board with the current Waku request path\n- map gateways, working memory, loop/tools, memory lifecycle, and eval/LLM-ops feedback to real modules\n- add a deterministic Excalidraw builder and SVG proxy for reproducible rendering and visual QA\n\n## Verification\n\n- Excalidraw binding validation: 98 elements\n- PNG render: 2520x1390, visually inspected\n- ruff: passed\n- release gate: 274 passed, 11 skipped; judge skipped because no API key was configured


ShenSeanChen:
the generator is the right instinct and it is the part i want to keep. i ran `python -m waku.ops.whiteboard.build_waku_architecture` and it reproduces the committed `.excalidraw` byte for byte, so the diagram is code now instead of a drawing that rots. that is the whole problem with the old one. the png going from 1.9mb to 380kb is a nice bonus, and the failure invariant line at the bottom is a genuinely good call, it is the exact contract our tests pin.

but i cannot merge this layout yet. when i actually look at the board rather than the diff, it is hard to read:

1. the memory section is spaghetti. four arrows cross in the same small area and the labels land on top of things. "facts" is printed over the EPISODIC box, and the two "retrieve" labels almost sit on each other.
2. three long diagonals cut across the whole board. PROCEDURAL and RETRIEVAL GATE fire up-right into WORKING MEMORY, FINAL REPLY fires down-left into ONE EVENT STREAM. each one crosses a section frame and other boxes on the way.
3. labels sit on the section borders. "message first" and "matched into prompt" both straddle the red HARNESS frame.
4. the space is unbalanced. top left and bottom right are empty while middle left is crammed.
5. the LLM ellipse is the only ellipse on a board of rounded rectangles.
6. overall it reads like graphviz output rather than the hand drawn style the other boards in `docs/whiteboards/` use.

what i would do: keep memory flowing left to right so the arrows do not have to double back, route the cross-section arrows around the frames instead of through them, and give every label its own clear space. `docs/whiteboards/k3-architecture.excalidraw` and `pi-architecture.excalidraw` are the style reference.

also missing from the board: the Arena (model racing, `waku/ops/arena.py`), google calendar, and whatsapp. arena especially, it is a headline feature.

i know that is a lot of notes on something you clearly put real work into. the generator idea is good enough that i would rather you iterate on it than have me redraw it, so take another pass whenever you have time and i will look again quickly.

ShenSeanChen:
Merged — thank you Allen, and sorry this sat for a week.

I want to say why it was worth the wait, because the reason is better than "it looks nicer."

The chart on main was **describing the wrong system.** It still said *Hermes Agentic Tools*, *Hermes Agent (auxiliary models)*, *LLM Ops (Not Hermes)* — a leftover from when it was reconstructed from a PNG after the original was lost. The flagship diagram in the gallery was naming a different project. Yours has the real file paths: `waku/app.py`, `runtime/session.py`, `loop/agent.py`, `memory/`, `ops/`, `evals/`. That's the fix, and it's not cosmetic.

You also caught something I hadn't: the old chart was on legacy Virgil (`fontFamily 1`) while the house style is Excalifont (`5`). Yours is correct.

Two things I'll tidy up myself rather than send back to you, since they're taste calls on my own visual language:

- the arrows from working memory down to the three memory pillars cross most of the canvas
- boxes *inside* a coloured boundary should stay ink — the red/orange/blue/green belongs to the section frames, not the contents

Neither is worth another round trip. An accurate chart with imperfect routing beats a tidy chart describing the wrong software.

Thanks for reading the actual code to build this — the "failure invariant: gate fails open" note tells me you did.

PR #37, opened 2026-07-27
Add OpenCode Zen and Go providers

vcavichini:
Add two new OpenAI-compatible LLM providers from OpenCode (opencode.ai):

- opencode_zen: https://opencode.ai/zen/v1 — default model deepseek-v4-flash-free
- opencode_go: https://opencode.ai/zen/go/v1 — default model deepseek-v4-flash

Both reuse the existing OpenAICompatClient (~60-line adapter that translates
Anthropic Messages shape to OpenAI chat.completions wire format). No new
dependencies, no new wire code.

Changes:
- waku/loop/models.py: register both providers in PROVIDERS dict
- waku/ops/pricing.py: add PRICING, MODEL_PRICING, MODEL_CUTOFF entries
- waku/ops/catalog.py: set a browser-like User-Agent for model listing
  (opencode.ai blocks Python-urllib/3.x with 403 error code 1010)
- waku/ops/coding_eval.py: add PI_PROVIDER mapping for pi sub-agent
- .env.example: document OPENCODE_ZEN_API_KEY and OPENCODE_GO_API_KEY
- README.md: list both in the provider intro paragraph
- evals/deterministic/test_pinned_models.py: add new keys to PROVIDER_KEYS
  fixture + track WAKU_PROVIDER to prevent env leak into test_tool_trigger

Tested: 275 offline deterministic evals pass, 40 provider tests pass,
real chat completion verified on opencode_go endpoint.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,6 @@
 # ── Provider ────────────────────────────────────────────────────────────────
 # Pick ONE provider and paste its key below. That's the whole setup.
-#   anthropic (default) · openai · gemini · deepseek · minimax · kimi · glm · openrouter
+#   anthropic (default) · openai · gemini · deepseek · minimax · kimi · glm · openrouter · opencode_zen · opencode_go
 # No paid key? openrouter's default models are $0 ":free" ids (rate-limited).
 WAKU_PROVIDER=anthropic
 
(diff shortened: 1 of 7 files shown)


ShenSeanChen:
merged, thank you. this is the shape i want provider PRs to have: two `PROVIDERS` rows and nothing else, no new code paths, so it costs nothing to anybody who does not use it.

what i checked before merging:

- 280 deterministic tests pass, ruff clean, based on today's main
- both endpoints are real. `GET /zen/v1/models` returns 60 models, `/zen/go/v1/models` returns 23
- your default models actually exist in those catalogs, so neither one 404s on first run. that is the bit that usually breaks
- both providers show up in the dashboard settings picker and the model list fills from the live catalog

nice catch on the test isolation bug too. `apply_settings` writes straight to `os.environ` past monkeypatch, so `WAKU_PROVIDER` was leaking into `test_tool_trigger` and would have bitten somebody eventually. that was not your job to fix and you fixed it anyway.

not tested: an actual chat turn, i do not have an opencode key. if you have run one, drop a screenshot in the thread, would be good to have it on the record.

vcavichini:
Nice that you merged my PR and I'm happy to contribute.
Opencode zen has some free models, like deepseek-v4-flash-free.

PR #38, opened 2026-07-28
feat(gcal): replace gcloud ADC with InstalledAppFlow OAuth

### LINKED ISSUE #31: Google Calendar: replace gcloud ADC with a real in-app OAuth flow
`sync_to_google_calendar` (added in #27) authenticates with **Application Default Credentials** — `google.auth.default()`, which reads whatever `gcloud auth application-default login` left on the machine.

That's the server/developer pattern. For waku it means a user has to: install the gcloud CLI, create a Google Cloud project, enable the Calendar API, create an OAuth client, download `credentials.json`, and run a CLI command. Roughly 15 minutes, and most people will stop before the end.

Waku is a personal assistant people run on their own laptop. Connecting a calendar should feel like every other app: click connect, browser opens, click Allow, done.

## What to build

Swap ADC for `InstalledAppFlow` (from `google-auth-oauthlib`):

- on first use, open the browser to Google's consent screen
- cache the resulting token in `.waku/` (gitignored, same place the rest of the runtime state lives)
- refresh it automatically; only re-prompt when the refresh token is actually dead
- if there's no `credentials.json` yet, fail with an honest message that says exactly where to get one — same honesty contract as every other tool

Everything else about #27 stays: opt-in behind `WAKU_GOOGLE_CALENDAR=1`, dep behind the `[gcal]` extra, local SQLite/ICS remains the source of truth, `sendUpdates=none`.

## Verify

- a deterministic eval in `evals/deterministic/` with the flow mocked — no network, no browser
- token cached under `.waku/`, never in the repo
- the existing `test_google_calendar.py` still passes unchanged

## Out of scope

Shipping a waku-owned OAuth client ID so users skip `credentials.json` entirely. That needs an app registration and Google's verification process — worth discussing separately, not part of this.

---
**Comment here before you start and I will assign it to you**, so two people never build the same thing.

Tian-Timm:
Closes #31

## Summary of Changes
- **Dependencies**: Added \google-auth-oauthlib>=1.0\ under the \gcal\ extra in \pyproject.toml\.
- **OAuth Authentication**: Replaced \google.auth.default()\ with \InstalledAppFlow.from_client_secrets_file\ for in-app browser consent flow.
- **Token Caching & Auto-Refresh**: Saved authorization token to \.waku/token.json\ (gitignored). Refreshes token automatically when expired.
- **Honest Error Handling**: If \.waku/credentials.json\ is missing, returns an explicit error explaining where to obtain OAuth Client ID credentials.
- **Backward Compatibility & Testing**: Added \home: Path | None = None\ parameter to \sync_to_google_calendar\. Retained \google.auth.default()\ fallback when \home=None\ so existing \	est_google_calendar.py\ passes completely unchanged. Added new deterministic tests in \evals/deterministic/test_gcal_oauth.py\.

### DIFF
--- a/evals/deterministic/test_gcal_oauth.py
+++ b/evals/deterministic/test_gcal_oauth.py
@@ -1,0 +1,207 @@
+"""DETERMINISTIC EVAL — Google Calendar InstalledAppFlow OAuth flow."""
+
+from __future__ import annotations
+
+import sys
+from types import ModuleType
+
+from waku.tools import calendar
+
+
+def _install_fake_oauth_modules(monkeypatch, *, execute_error: Exception | None = None):
+    captured = {}
+
+    google = ModuleType("google")
+
+    # google.auth
+    google_auth = ModuleType("google.auth")
+    google_auth_requests = ModuleType("google.auth.transport.requests")
+
+    class Request:
+        pass
+
+    google_auth_requests.Request = Request
+    google_auth.transport = ModuleType("google.auth.transport")
+    google_auth.transport.requests = google_auth_requests
+    google.auth = google_auth
+
+    # google.oauth2.credentials
+    google_oauth2 = ModuleType("google.oauth2")
+    google_oauth2_credentials = ModuleType("google.oauth2.credentials")
+
+    class FakeCredentials:
+        def __init__(self, valid=True, expired=False, refresh_token="ref-token"):
+            self.valid = valid
+            self.expired = expired
+            self.refresh_token = refresh_token
+            self.refreshed = False
+
+        @classmethod
+        def from_authorized_user_file(cls, path, scopes=None):
+            captured["loaded_token_path"] = path
+            captured["loaded_scopes"] = scopes
+            return FakeCredentials(valid=captured.get("creds_valid", True), expired=captured.get("creds_expired", False))
+
+        def refresh(self, request):
+            captured["refreshed"] = True
+            self.valid = True
+            self.expired = False
+
+        def to_json(self):
+            return '{"token": "fake-oauth-token"}'
+
+    google_oauth2_credentials.Credentials = FakeCredentials
+    google_oauth2.credentials = google_oauth2_credentials
+    google.oauth2 = google_oauth2
+
+    # google_auth_oauthlib.flow
+    google_auth_oauthlib = ModuleType("google_auth_oauthlib")
+    google_auth_oauthlib_flow = ModuleType("google_auth_oauthlib.flow")
+
+    class FakeInstalledAppFlow:
+        @classmethod
+        def from_client_secrets_file(cls, path, scopes=None):
+            captured["client_secrets_path"] = path
+            captured["flow_scopes"] = scopes
+            return FakeInstalledAppFlow()
+
+        def run_local_server(self, port=0):
+            captured["ran_local_server_port"] = port
+            return FakeCredentials()
+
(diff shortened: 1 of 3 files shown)


ShenSeanChen:
tested this properly. 4/4 your own tests, 284 across the suite, ruff clean, and the dep is correctly behind the `[gcal]` extra. the token caching, the refresh-if-expired path and the fallback to a repo-root `credentials.json` are all right. thank you.

and `flow.run_local_server(port=0)` is exactly the right call — that IS the embedded popup i wanted. browser opens, user approves, redirect to localhost, no server anywhere.

but i think we fixed the wrong layer, and that is on me for how i wrote the issue.

the problem was never the auth mechanism. it is **who registers the app with google**. right now every user has to go make their own google cloud project, enable the api, create an oauth client, download a json, and put it in the right folder. that is five steps before waku can read one event. most people will just not.

what i actually want:

1. **waku ships its own client id.** i register one google cloud project for the whole repo, the client id and secret live in the code, and a user just clicks "connect google calendar". this is what gcloud, rclone and basically every cli tool does — for "desktop app" oauth clients google explicitly accepts that the secret cannot be kept confidential.
2. **keep `credentials.json` as the escape hatch** for anyone who would rather use their own project. your code already reads it, so that path stays exactly as you built it.
3. **switch the default scope to `calendar.events.readonly`.** right now we ask for `calendar.events`, which is read AND write, and google treats calendar write as a sensitive scope — that is the slow verification tier. reading is what people actually want first. keep the write scope behind an opt-in for people who want waku creating events.

two honest caveats so you are not surprised: until the app passes google verification users see an "unverified app" screen and there is a 100 user cap, and github secret scanning may flag the shipped client secret even though google does not consider it one.

i am registering the google cloud project now and will drop the client id in this thread. if you want to take the change, it is roughly: scope constant split into read/write, client id/secret constants with the `credentials.json` path kept as an override, and a test that the readonly scope is the default.

no rush and no obligation — if you would rather i finish it on top of your commit, say so and i will, with your commit kept in the history. you have done the hard part.

ShenSeanChen:
merged. thank you, and sorry for the wait.

i want to correct something i said earlier in this thread. i worried this was overlapping with work i did on the read side, so i went and checked properly: `waku/tools/calendar.py` on main had zero `InstalledAppFlow` before your commit. what i shipped was a separate file for READING google calendar. yours fixes the auth on the WRITE path. they do not overlap at all, and i should have checked before saying otherwise.

what i verified before merging:

- 4/4 your own tests, and 284 across the suite
- ruff clean, and the dep correctly behind the `[gcal]` extra
- the token caching, the refresh-if-expired path, and the fallback to a repo-root `credentials.json` all read correctly
- `flow.run_local_server(port=0)` is the right call and i confirmed it live on my own machine — browser opens, approve, redirect to localhost, no server anywhere. that is exactly the embedded sign-in i was asking for.

the client-id idea from my earlier comment is still worth doing, but it is a separate change and not a condition on yours. i left the stubs in `waku/tools/google_calendar.py` for whoever picks it up — you would be very welcome to.

that is 6 merged now, more than anyone except me. thank you for keeping at it.

Tian-Timm:
Thanks for merging!

The client-id/default readonly scope improvement makes sense. I'll take this follow-up and open a new branch.

I'll keep `credentials.json` as an override, add the default client config, split read/write scopes, and add tests for the default behavior.

PR #42, opened 2026-07-31
feat(gcal): add bundled OAuth client support

Tian-Timm:
## Summary

- Add bundled OAuth client support for Google Calendar
- Keep `credentials.json` as an OAuth client override
- Default OAuth scope is now `calendar.events.readonly`
- Unify OAuth flow and token loading scopes
- Add tests for bundled client fallback and override behavior

## Testing

- Google Calendar OAuth tests passed
- Ruff check passed

Note:
The bundled OAuth client config is currently a placeholder.
The project-owned Google OAuth client ID/secret can be added once available.


ShenSeanChen:
Merged — thanks Timm, that's seven now.

Tested before merging: full suite green on the branch (295), your 8 new tests, ruff clean, and `list_google_events` against a real cached token still reads a live calendar correctly.

One thing I want to record because it's the best part of this PR and easy to miss: dropping `WRITE_SCOPE` isn't a feature removal. The write path (`create_event` syncing to Google) uses a **separate** token file and its own flow, so that scope was being requested and never exercised — waku was asking users for calendar-write permission it never used. Narrowing to read-only is strictly better.

Follow-up for whoever wants it: two token files for one Google account (`token.json` for writes, `google-token.json` for reads) means two consent screens. Worth unifying.

Next step is mine — filling in the real client ID and secret.

PR #58, opened 2026-08-01
feat(dashboard): copy buttons and fenced code blocks

### LINKED ISSUE #59: Dashboard chat: fenced code blocks render as plain text, no copy on replies
## Problem

When the agent replies with fenced markdown code blocks (triple backticks), the dashboard chat renders them as **plain text** — backticks, language tags, and code all show inline instead of as a styled code block. There is also no way to copy a full reply or a code snippet in one click.

## Screenshot (before)

Dashboard renders fenced code as plain text

In the example above, a Python `two_sum` solution should appear as a fenced code block with monospace formatting; instead the raw ``` markers and code run together as body text.

## Expected

- Fenced code blocks render with monospace styling, optional language label, and a **Copy** button
- Horizontal rules (`---`) render as `<hr>`
- Chat reply cards expose a hover **Copy** button for the full assistant message

## Fix

Addressed in #58.

kirtanchandak:
**What this does, and why**

Dashboard chat replies and markdown code fences get one-click copy. The lightweight markdown renderer also gains fenced code blocks (with optional language label) and horizontal rules — replies that include ``` blocks or `---` no longer render as plain text.

**Which issue does it close?** Closes #59

## Before / After

| Before | After |
|--------|-------|
| Fenced code renders as plain text — backticks and code run inline | Styled code block with language label + **Copy** button; reply cards get hover **Copy** |

**Before**

Before — fenced code renders as plain text

**After**

After — fenced code block with copy button

---

- [x] **A deterministic eval** in `evals/deterministic/` covers the behavior.
- [x] `make gate` and `make lint` pass locally.
- [x] Any heavy or optional dependency is **behind an extra**, not in the default install.
- [x] No hidden network calls, no reading secrets or `.env`, nothing runs at install time.

```
make dashboard
Sent a reply containing a fenced code block and --- ; copy buttons on reply cards and code fences work; hr renders correctly.
```

**Anything you're unsure about?** Static JS/CSS only — no Python path changes.

---
First PR here? CI needs a maintainer to approve your workflow run. If it looks stuck, say so on the PR — that delay is ours, not yours.


ShenSeanChen:
Merged — thanks Kirtan, and thanks for filing #59 before fixing it. That's the right order and not everyone does it.

One thing I changed straight after, and I want to be open about it rather than have you find it in the log.

The copy button puts the reply into an HTML attribute:

```js
<button class="msg-copy" data-text="${esc(t.reply)}">
```

waku's `esc()` escaped `&`, `<` and `>` — but not quotes. So a reply containing one double quote closed the attribute:

```
data-text="hi" onmouseover="alert(1)" x=""
```

That matters here because the model's output isn't entirely ours — `search_web` and `browse_web` pull text off the open web — and this dashboard has the memory and the settings in it.

**This is not a mistake on your part.** `esc()` was incomplete, and nothing had put model text in an attribute before, so nothing could reveal it. Your PR is what made it visible. Fixed at the helper in #69 with tests, so every future caller inherits it.

Worth noting the other half held up perfectly: your fenced code blocks are safe, because `renderMarkdown` escapes the whole reply before parsing — I fed it a `<script>` tag inside a fence and it came out as `&lt;script&gt;`.

Good first contribution. If you want another, #66 and #67 are open — tools with no test coverage.

Issue #32, opened 2026-07-26
A Connections surface: one registry for every integration, instead of hand-editing .env


ShenSeanChen:
Waku can connect to about ten external things — Telegram, Discord, WhatsApp (#26), Notion, Supabase, Google Calendar (#27), Apple Calendar/Mail, OpenTelemetry, Tavily — plus nine model providers. Every one of them is configured the same way: read the README, find the right variable name, hand-edit `.env`, restart.

There is no single place that answers **"what am I connected to, and what's broken?"** The dashboard surfaces status for exactly five of them, hand-written one field at a time in `settings_info()`. Everything else you discover by grepping `.env.example`.

That's the gap. It also gets worse with every integration merged, and we merged three this week.

## The shape

One declarative registry, three surfaces reading from it.

```python
# waku/integrations.py
@dataclass(frozen=True)
class Integration:
    key: str                      # "google_calendar"
    name: str                     # "Google Calendar"
    what: str                     # one line: what it gives the agent
    env: list[str]                # env vars it reads
    extra: str | None             # pip extra, e.g. "gcal"
    is_connected: Callable[[Settings], bool]
    setup_url: str                # where to get the credential
```

Then:

- **Dashboard → a Connections tab.** Every integration as a row: name, one-line description, a status pill (connected / not configured / installed-but-unconfigured / error), and the fields to fill in. Same in-process rebuild the model picker already uses, so no restart.
- **CLI → `waku connections`.** Same list in the terminal, for people who never open the dashboard.
- **`.env.example`** generated from the registry, so it can't drift from the code again.

## Why a registry rather than more fields in settings_info

`settings_info()` has five hand-written status fields today, and adding an integration means editing `dashboard.py`, `.env.example`, `config.py` and the README, hoping you got all four. This repo has already been bitten twice by exactly that class of drift. One source of truth, three readers.

It's also the natural home for #31: "Connect Google Calendar" becomes a button that runs the OAuth flow, not a CLI incantation.

## Verify

- a deterministic eval asserting **every** integration in the registry has a status function and that no env var in `.env.example` is missing from it (the anti-drift test)
- dashboard smoke test: an unconfigured integration says so honestly, a broken credential shows the error rather than failing silently
- no new core dependency; heavy ones stay behind their existing extras

## Not in scope

Actually implementing the OAuth flows — that's #31 and its siblings. This issue is the surface they plug into.

---
**Comment here before you start and I will assign it to you**, so two people never build the same thing.

[...]

PR #72, opened 2026-08-03
feat: unified Connections registry, provider grid, and gateway lifecycle

wilenWang:
Issues: #32 

## Summary
This PR introduces a single, declarative registry for every user-configurable connection and provider, replaces the monolithic Settings form with dedicated Models and Connections pages, and makes Telegram/Discord/WhatsApp gateways reloadable at runtime without restarting the dashboard.
(PS: I’ve spent hours digging into different connectors and getting them connected.😵‍💫)

#### Model
image

#### Connection
image

## Key changes:

1. Centralized integration registry (waku/integrations.py)
• New Integration / Provider dataclasses describe env vars, install commands, setup URLs, reload behavior, and health probes.
• All connections (Telegram, Discord, WhatsApp, Google Calendar, Apple Calendar, Apple Tools, Notion, Supabase, Tavily, OpenTelemetry) and all AI providers are declared in one place.
• apply_integration, apply_provider, test_integration, etc. own read/write/probe logic so settings_api.py no longer needs to know about every backend.
• .env.example is now auto-generated from the registry via scripts/generate_env_example.py, keeping comments and examples in sync.

2. New dashboard pages
   • Models (#models): a card grid of AI providers with logo, status dot, enable/disable toggle, and an edit modal for keys / endpoint / model
     defaults.
   • Connections (#connections): a card grid of integrations grouped by category, with per-connection config modals, test buttons, and force-save.
   • Provider endpoint selection is now first-class (e.g., MiniMax China/Global, Moonshot Global/China, Zhipu Global/China).

3. Gateway lifecycle management
   • GatewaySupervisor (waku/gateway/supervisor.py) starts, stops, and restarts gateways when their config fingerprint changes.
   • GatewayAgentRunner (waku/gateway/runner.py) runs the synchronous Waku agent on a dedicated worker thread per async gateway, keeping SQLite access single-threaded.
   • Telegram, Discord, and WhatsApp adapters now return a GatewayHandle with stop() / status() so the supervisor can manage them uniformly.
   • Dashboard main() now reconciles gateways on startup and shuts them down cleanly on exit.

4. Health probes
   • Added read-only connection health checks for Apple Calendar, Apple Tools, Google Calendar, and OpenTelemetry.
   • test_integration exposes probe results to the dashboard UI.

5. Provider model picker improvements
   • Searchable model picker with live catalog fetching.
   • Provider enable/disable toggle split from "configured" status.
   • disabled_providers persisted in settings so users can hide providers from the chat switcher without deleting keys.

6. Tests
Added or expanded deterministic coverage:
   • test_integrations.py — registry round-trips, apply/test flows
   • test_gateway_supervisor.py / test_gateway_runner.py — lifecycle and threading
   • test_dashboard_routes.py — new /api/connections, /api/connections/test, /api/providers
   • test_provider_disabled.py / test_provider_base_urls.py — provider grid behavior
   • test_apple_calendar.py / test_apple_tools.py — macOS health probes
   • test_gcal_oauth.py / test_connections_cli.py / test_static_assets.py

## Migration / breaking notes

   • The old /api/settings write path no longer accepts keys, provider, model, episodic_store, etc. Those are now handled by /api/providers and
     /api/connections.
   • .env.example is generated; hand-edits should go into waku/integrations.py or scripts/generate_env_example.py.

## What I could NOT manually verify
1. **WhatsApp gateway**: I cannot create a Meta Business Portfolio, so I cannot create a WhatsApp app or link a phone number to test the Cloud API webhook end-to-end.
2. **Model providers beyond DeepSeek, Kimi, and MiniMax**: I only have API keys for those three; other providers were not manually tested.
3. **Supabase memory embedding**: this path depends on an OpenAI API key for the embedding model, which I do not have, so embedding-based memory tests/verification are missing.
Anybody can work on this branch to help~
All other connections were manually tested.


ShenSeanChen:
@wilenWang — thank you, and sorry this sat for a week. This is the largest contribution the project has had, and it builds the thing I designed in #32 and then never got to. It deserved a faster reply than it got.

I ran it properly before writing this: fresh worktree, no `.env`, dashboard on a second port so I was seeing exactly what a stranger sees after `pip install`. **489 passed, ruff clean, zero console errors.** Both pages look right — grouped by category, honest status dots, the house style held. Two things I specifically went looking for and was glad to find: the new JS routes user strings through `esc()` (we had an attribute-escape bug in #69, so that mattered), and `browser_agent.rebuild()` plus the `tracer.event("config", ...)` audit record both survived the move into `integrations.py`, so a brain swap is still traceable.

Three changes and it goes in.

**1. Delete the dead code in `views.js`.**

There are ~120 lines of the old Settings markup after an unconditional `return`, with a comment saying it is "intentionally unreachable while retained temporarily as a reference." I understand the instinct, but nothing in this repo lints JS, so it will never be caught and never be removed. It also reads fields the API no longer sends (`p.default_model`, `p.key_set`), so the next person to open that file learns a shape that no longer exists. Git remembers it — please just delete it.

**2. Wire the `.env.example` drift check into CI.**

`scripts/generate_env_example.py --check` works — I ran it, exit 0. But it is connected to nothing: not the workflow, not `make gate`, not a test. The first person who adds an integration and forgets to run `--write` will silently ship a stale `.env.example`, and the whole point of generating it is that it cannot drift. One step in `.github/workflows/` closes that.

**3. Drop the three whiteboard files.**

`waku/ops/whiteboard/style.py`, `build_k3_tutorial.py` and `build_pi_system.py` are pure lint churn here (`lambda` -> `def`, splitting semicolon statements). Nothing to do with Connections, and it makes the diff harder to review. Happy to take them as their own PR if your linter is flagging them.

One thing I am **not** blocking on, just so you know it exists: `settings_info()` lost its `or prov.model` fallback, so on a keyless install the nav renders `anthropic ·` with a trailing separator and no model name. The agent is unaffected — `waku/loop/models.py:227` still resolves the provider default at build time — so it is display-only. Fix it here if it is quick, otherwise I will.

On the parts you could not verify: WhatsApp, the providers you have no keys for, and Supabase embeddings. That is the right call and thank you for saying so plainly instead of implying coverage you did not have. I will exercise those myself before this lands, and I will not hold the PR waiting for you to buy API keys.

Ping me when the three are pushed and I will run it again.

ShenSeanChen:
@wilenWang — ignore the three asks in my last comment, I have pushed them to your branch myself. You built the thing; you should not also have to do the tidying. Three commits, all small:

- `bc3d40f` the frontend fixes (see below — one of them is not a nit)
- `da11f9c` the `.env.example` drift check, one step in the workflow
- `ea01db6` the three whiteboard files reverted to main

**The one that matters: the graph-workflows toggle had disappeared.**

Not broken — absent. `saveSettings()` still reads `#set-graph-workflows`, but the new Settings view no longer renders it, so the lookup returned `undefined`, `apply_settings` read that as "don't touch", and the flag became unwritable from the browser. Nothing threw. Meanwhile the Overview panel still tells the user, in those words, to "switch on **graph workflows** in Settings" — pointing at a control that was not there.

That flag is the door to the whole graph pillar: when it is on, every message is triaged through a graph before the loop. Restored verbatim from main and verified end to end — set it on, save, `WAKU_GRAPH_WORKFLOWS=1` lands in `.env` and the API reports it back.

I only found it because I ran your branch with **no `.env` at all**, which is the fresh-install case neither of us sees day to day. Worth stealing as a habit: `WAKU_DASHBOARD_PORT=7778` in a worktree with no env file shows you exactly what a stranger meets after `pip install`.

Two smaller ones in the same commit: `settings_info()` had lost its `or prov.model` fallback, so a keyless install rendered `anthropic ·` in the nav — a trailing separator with no model after it. And the Arena's empty state said "add some in Settings", which was true before pinning moved to the Models grid; it now links to Models.

For the record, I got the model fix wrong on the first attempt — a walrus inside an `or`, which short-circuits, so the variable never binds and the next line is a `NameError` as soon as `WAKU_MODEL` is set. Your branch was 489 green and mine was 487 until I rewrote it plainly. Your tests caught me, which is a good sign about the tests.

Current state: **489 passed, ruff clean, zero console errors**, both pages verified in a browser in light and dark. Sean is doing the final click-through himself — the dashboard is what he films, so a person looks at it before it lands, not just a test suite. Nothing needed from you unless you disagree with any of the three.

Thank you again for this. It is the biggest thing anyone has contributed and it closes an issue I wrote and then left sitting for a month.

wilenWang:
> @wilenWang — ignore the three asks in my last comment, I have pushed them to your branch myself. You built the thing; you should not also have to do the tidying. Three commits, all small:
> 
> * `bc3d40f` the frontend fixes (see below — one of them is not a nit)
> * `da11f9c` the `.env.example` drift check, one step in the workflow
> * `ea01db6` the three whiteboard files reverted to main
> 
> **The one that matters: the graph-workflows toggle had disappeared.**
> 
> Not broken — absent. `saveSettings()` still reads `#set-graph-workflows`, but the new Settings view no longer renders it, so the lookup returned `undefined`, `apply_settings` read that as "don't touch", and the flag became unwritable from the browser. Nothing threw. Meanwhile the Overview panel still tells the user, in those words, to "switch on **graph workflows** in Settings" — pointing at a control that was not there.
> 
> That flag is the door to the whole graph pillar: when it is on, every message is triaged through a graph before the loop. Restored verbatim from main and verified end to end — set it on, save, `WAKU_GRAPH_WORKFLOWS=1` lands in `.env` and the API reports it back.
> 
> I only found it because I ran your branch with **no `.env` at all**, which is the fresh-install case neither of us sees day to day. Worth stealing as a habit: `WAKU_DASHBOARD_PORT=7778` in a worktree with no env file shows you exactly what a stranger meets after `pip install`.
> 
> Two smaller ones in the same commit: `settings_info()` had lost its `or prov.model` fallback, so a keyless install rendered `anthropic ·` in the nav — a trailing separator with no model after it. And the Arena's empty state said "add some in Settings", which was true before pinning moved to the Models grid; it now links to Models.
> 
> For the record, I got the model fix wrong on the first attempt — a walrus inside an `or`, which short-circuits, so the variable never binds and the next line is a `NameError` as soon as `WAKU_MODEL` is set. Your branch was 489 green and mine was 487 until I rewrote it plainly. Your tests caught me, which is a good sign about the tests.
> 
> Current state: **489 passed, ruff clean, zero console errors**, both pages verified in a browser in light and dark. Sean is doing the final click-through himself — the dashboard is what he films, so a person looks at it before it lands, not just a test suite. Nothing needed from you unless you disagree with any of the three.
> 
> Thank you again for this. It is the biggest thing anyone has contributed and it closes an issue I wrote and then left sitting for a month.

All good on my side — thanks for jumping in and taking care of those, especially catching the missing graph-workflows toggle. The three commits look good to me.

PR #81, opened 2026-08-07
Add scored eval for retrieval gate accuracy (#77)

### LINKED ISSUE #77: The retrieval gate's accuracy has never been measured
The retrieval gate is waku's headline memory idea: before every turn, a cheap
model decides whether to search memory at all, on the theory that irrelevant
memories bias the answer more than missing ones do.

We have never measured whether it decides correctly.

`evals/deterministic/test_retrieval_gate.py` has eleven tests. Every one of them
is about plumbing:

- does it parse JSON out of surrounding prose
- does it survive a reasoning block before the JSON
- does it fail open on malformed JSON / an API error
- is it exactly one model call

Not one of them asks whether a "yes" was the right answer. So the claim the
whole design rests on — that gating saves work without losing recall — is
asserted, not evidenced. Simon Strandgaard's agent-memory atlas review (#45)
made exactly this point, and it is correct.

## What this needs

A labeled dataset plus a scored eval, in the style of `evals/judge/`, not
`evals/deterministic/` — this is a quality judgment, not a 0/1 assertion.

1. A dataset of messages paired with the memory that exists at the time, each
   labeled `should_retrieve: true | false`. Cover the cases that actually
   matter:
   - chitchat with a full memory store (should NOT retrieve)
   - a direct question about a stored fact (SHOULD)
   - a follow-up whose referent is only in history, not memory (should NOT)
   - a question whose answer memory does not contain (SHOULD retrieve, and
     then find nothing — the gate's job is not to predict success)
2. Report both error directions separately. They are not symmetric: a false
   "no" silently loses a fact the user told us, a false "yes" costs one search
   and some prompt space. The first is much worse and the headline number must
   not hide it.
3. Score the generated `query` too, not just the boolean. A correct "yes" with
   a useless query retrieves nothing and looks identical to a correct "no" from
   the outside.

## Why it matters beyond the gate

The same measurement gap applies to the cost claim in the README. We say the
gate saves work; we have never published a number for how much, on what
traffic. A dataset here gives us both.

## Where to start

- `waku/memory/retrieval_gate.py` — the 55 lines under test
- `evals/judge/test_response_quality.py` — the scored-eval pattern to copy
- `evals/deterministic/test_retrieval_gate.py` — what exists today, and stays;
  this issue adds a sibling, it does not replace plumbing tests

YashSensei:
## What this adds

A scored eval that measures whether the retrieval gate decides correctly — the thing issue #77 says has never been measured. One new file: `evals/judge/test_retrieval_gate_accuracy.py`. The gate itself (`waku/memory/retrieval_gate.py`) is untouched; this is eval only.

Closes #77.

## The three things the issue asked for

**1. Labeled dataset + scored eval (`evals/judge/` style)**

12 cases across the four categories from the issue:

| Category | Example | Expected |
|---|---|---|
| chitchat (3 cases) | "Hey, how's it going?" | should NOT retrieve |
| direct-fact (3 cases) | "What's my dog's name?" | SHOULD retrieve |
| followup-history (3 cases) | "Can you make it earlier?" | should NOT retrieve |
| missing-memory (3 cases) | "What did I have for breakfast?" | SHOULD retrieve |

DeepEval GEval, scored 0-1 at threshold 0.6 — not a 0/1 assertion. The judge sees the message + memory context + the gate's decision, but **never** the expected label.

**2. Error directions reported separately, asymmetry not hidden**

The headline is a **cost-weighted score** (`FALSE_NEGATIVE_COST = 4.0`, `FALSE_POSITIVE_COST = 1.0`), not raw accuracy. A false NO (silently loses a fact the user told us) counts 4x a false YES (one local search + some prompt space). Raw accuracy prints second, labelled as "treats both errors as equal, which they are not".

Both directions get their own headed section with rates. Any false negative prints a `*** WARNING ***` block naming the missed case and quoting the gate's own reason.

The soft floor asserts on the weighted score, not raw accuracy — a gate that gets every skip-case right and every retrieve-case wrong is useless, and only the weighted number says so.

**3. Query quality scored, not just the boolean**

The gate returns `(retrieve, query, reason)`. A correct "yes" with a useless query retrieves nothing and looks identical to a correct "no" from the outside. `GateQueryQuality` scores the generated query: told the target is a keyword index (SQLite FTS5), told not to penalise for empty stores (coverage is a store property, not a query property), told not to score it as an answer.

Two skip paths: gate said no (no query to score), gate said yes on a skip-case (the decision eval owns that failure — scoring the query would report the same mistake twice).

## Design decisions worth flagging

**`evaluation_steps` instead of `criteria`** — `test_response_quality.py` uses `criteria` (GEval generates its own rubric). This file uses explicit `evaluation_steps` instead. Given `criteria`, the judge kept generating steps that contradicted the stated rules: consulting memory context it was told to ignore, and scoring a search query as if it were an answer ("Step 4's core requirement: it does not actually answer what the user had for breakfast" — a step that was never asked for). Explicit steps remove the generation step so the rules can't be paraphrased away between runs. Five consecutive clean runs after the switch. If the maintainer would rather both judge files look identical, the fallback is restricting query scoring to direct-fact cases (sidesteps the empty-store problem but halves coverage).

**Information horizon** — the gate sees only the message, not chat history or memory. The judge sees more. Without stating this in the evaluation steps, the judge scored "retrieve" on "Can you make it earlier?" as wrong because from the memory context it's obviously a self-contained follow-up. From the gate's vantage point it's a personal-sounding sentence with no context at all, and looking is the defensible move. The steps now explicitly state the information horizon.

## Measured results (live run with API key)

```
COST-WEIGHTED SCORE:   90.0%   <- headline: a false NO counts 4x a false YES
Accuracy (unweighted): 75.0%
  Missed retrievals (false NO):    0 of 6   miss rate 0.0%
  Needless retrievals (false YES): 3 of 6   false-alarm rate 50.0%
Query coverage: 6 of 6
```

All errors are in the cheap direction (false YES on followup-history cases). Zero false negatives — no fact the user told us was silently lost. That's the fail-open design working: the gate errs on the side of searching, which is exactly the asymmetry this eval exists to verify.

## Verification

- `ruff check` — clean
- `pytest --collect-only` — 25 items (12 decision + 12 query + 1 summary), no errors
- `pytest evals/deterministic -q` — 416 passed, 13 skipped (no regressions)
- Judge tests skip without an API key (`pytestmark`), so CI stays green without secrets


ShenSeanChen:
Merged — thanks @YashSensei, and sorry this sat for eight days.

Your numbers reproduced **exactly** on my machine, not approximately: 90.0% weighted, 75.0% raw, 0 of 6 missed retrievals, 3 of 6 needless, 6-of-6 query coverage. 19 passed, 6 skipped, 48s.

I also sabotaged the gate to never retrieve, to check the soft floor is real:

```
COST-WEIGHTED SCORE:   20.0%
Missed retrievals: 6 of 6 (100%)
Confusion: TP=0 FP=0 TN=6 FN=6
AssertionError: Cost-weighted gate score 20% is below 50%
```

That run is the best argument for your own design: **raw accuracy there is 50%, which looks survivable.** A gate that has silently stopped consulting memory at all scores an unremarkable-looking half, and only the weighted number calls it what it is. The asymmetry was not decoration.

Two other things worth naming, because they are the parts a reviewer notices and a drive-by contributor does not: stating the information horizon in the evaluation steps (the gate sees the message; the judge sees more), and skipping query scoring on a gate-said-yes skip-case so one mistake is not reported twice.

On `evaluation_steps` vs `criteria` — your reasoning holds and it stays as written. `test_response_quality.py` is the file that should move to match, not the other way round; I am opening an issue for that rather than dragging the better convention down to the older one.

Closes #77.

YashSensei:
Thanks for merging it — and for running the sabotage check. That 20% vs 50% gap is exactly why I built the eval around cost-weighted scoring: a totally broken gate can still look "okay" on raw accuracy, and the headline number has to surface the expensive direction.

Happy to keep contributing. If there are other gaps in the eval suite or the memory layer where measurement/testing would help, point me at them.

PR #92, opened 2026-08-08
memory/mem0 adapter

ShenSeanChen:
- **fix(ui): the Connections group is Memory, not Storage**
- **feat: Mem0 as a semantic-memory backend, behind the same six methods**


ShenSeanChen:
Superseded by #93, which landed the same adapter — `waku/memory/semantic/mem0_store.py` and the mem0 registry entries are already on `main`. That is why this branch now shows as conflicting: it conflicts with its own merged content. Closing rather than merging so the queue reflects reality.

PR #95, opened 2026-08-09
ui: thin dark scrollbars + color-scheme declaration

snikmas:
## What
- Declare `color-scheme: light dark` on `:root` so native scrollbars follow dark mode (they were rendering white/light against the dark UI)
- Thin (6px) theme-colored scrollbar styling via `scrollbar-width`/`scrollbar-color` (Firefox) and `::-webkit-scrollbar` (Chrome/Safari)

## Why
Native scrollbars stayed in OS light mode because the page only themed CSS variables under `prefers-color-scheme: dark` without declaring `color-scheme`. Result: bright white scrollbars over the dark console.
Before: 
a125501e83cea5504e8a3e4d5483c585
(Ubuntu)

Now: 
image


snikmas:
(If firstly have to open an issue and only after that can open a pull request - sorry, I can open an issue)

ShenSeanChen:
Merged — thanks @snikmas.

I measured both sides before merging rather than trusting a screenshot, since this is native OS rendering:

| | main | this PR |
|---|---|---|
| `color-scheme` | `normal` | `light dark` |
| `scrollbar-color` | `auto` | `rgb(51,51,56)` + transparent track |

Both properties genuinely apply — it is not a no-op.

Worth recording why this sat unreviewed: on macOS the measured scrollbar width is **0px on both sides**, because overlay scrollbars take no layout space. The bug is essentially invisible on the maintainer machine, which is exactly why your Ubuntu before/after was the useful part of the report. Thank you for including it.

The `color-scheme` declaration is the half that helps everyone — waku themed its own variables under `prefers-color-scheme: dark` but never told the browser the page supported dark mode, so every native control was left in light mode. That was a real gap and nobody had noticed.

PR #96, opened 2026-08-10
feat: strip delegate-dangerous env vars from all subprocess spawn points

Behruz44:
## Summary

delegate_task spawns pi as an unsandboxed subprocess with full filesystem and network access. workspace.autorun goes further - it executes arbitrary code written by pi. Any credential present in waku's environment is readable by both.

This PR adds WAKU_DELEGATE_ENV_DENY - an environment variable (comma-separated) that host applications can set to name vars stripped from every delegated subprocess. Applied at three spawn points: Popen in _run_pi_json, subprocess.run in delegate_task, and subprocess.run in workspace.autorun.

WAKU_DELEGATE_ENV_DENY itself is deliberately not stripped, so nested waku invocations inside the delegate inherit the same policy.

The env-stripping logic lives in a new shared module waku/tools/_env.py, imported by both experimental.py and workspace.py.

Not covered: the pi --help probe in _pi_supports_json() intentionally does not strip env - --help prints usage text and exits without executing user code, so the risk is minimal.

## Threat model

- Before: host app sets MY_SECRET=xxx in env, waku inherits it, pi can printenv MY_SECRET, leak. Worse: pi writes a script that reads os.environ, autorun executes it with the secret in env, leak.
- After: host app sets WAKU_DELEGATE_ENV_DENY=MY_SECRET, delegate_env() strips it at all three spawn points, neither pi nor pi-written code can read it.


ShenSeanChen:
Merged — thanks @Behruz44.

Two details in this design that most people would have missed, and they are why it went in as-is:

- the denylist is read **at call time**, not import time, so a host app can set the policy after importing waku
- `WAKU_DELEGATE_ENV_DENY` itself is deliberately not stripped, so a nested waku inside the delegate inherits the policy instead of silently losing it

I verified it end-to-end before merging, including the control case — a real child process spawned through `workspace.autorun`, printing a seeded secret when the policy is off and `<stripped>` when it is on. 8/8. Full deterministic suite 564 passed, ruff clean.

Two things I am pushing as a follow-up rather than asking you to round-trip:

**1. One spawn point is still uncovered.** `waku/ops/coding_eval.py` runs the same pi binary — its own docstring calls it "the same sub-agent `delegate_task` uses" — and passes `env=os.environ.copy()`, the exact line `delegate_env()` exists to replace. It then runs the `verify` command with `shell=True` in a directory full of code pi just wrote. Your threat model applies there word for word.

**2. No deterministic eval.** This is security code, and a denylist that quietly stops working looks exactly like one that works. I am turning my verification into `evals/deterministic/test_delegate_env.py`.

The design credit is yours; we will carry the coverage.

PR #100, opened 2026-08-11
test: add deterministic tests for list_events tool

### LINKED ISSUE #66: list_events has no test coverage — and it once reported an empty calendar that was full
## What

`list_events` (`waku/tools/calendar.py::make_list_tool`) has **zero test references** in `evals/`. It is the tool that answers "what's on my calendar", and it merges results across every connected source — Google Calendar and the local `.ics` — labelling each line by where it came from.

## Why it matters

This is not hypothetical. On 2026-07-30 it reported an empty calendar while the real day was full, and nothing in CI noticed, because nothing tests it.

## What a good fix looks like

Add `evals/deterministic/test_list_events.py`. It should run **offline** — no network, no Google credentials, no real calendar. Follow the pattern in `evals/deterministic/test_apple_tools.py`: assert on the logic you *can* check without the outside world.

Worth pinning:

- events from two sources are merged into one list, each labelled with its source
- an empty result names the calendars that **were** checked, rather than just saying "nothing" — silence and "nothing found" are different answers and the difference matters
- one source failing (no Google credentials, unreadable `.ics`) still returns the other source's events rather than an error
- the date window is respected at the boundaries

## Where to look

- `waku/tools/calendar.py` — `make_list_tool(conn, home)`
- `evals/deterministic/test_apple_tools.py` — the closest existing example of testing a tool whose real dependency can't run in CI
- `CONTRIBUTING.md` for the house style

Run the suite with:

```bash
WAKU_PROVIDER=anthropic ANTHROPIC_API_KEY= .venv/bin/python -m pytest evals/deterministic -q
```

Comment here and I'll assign it to you.

kirtanchandak:
Resolves #66. Adds offline deterministic tests for the `list_events` tool.


ShenSeanChen:
Merged — thanks @kirtanchandak.

Before merging I did the thing that actually proves a test suite is real: broke `list_events` four ways and checked yours goes red each time.

| what I broke | your tests that caught it |
|---|---|
| stopped reading Google entirely — the exact #66 bug shape | **2 failed** |
| dropped the `Checked:` list, back to a bare "no events" | **1 failed** |
| treated a Google *failure string* as if it were data | **2 failed** |
| off-by-one on the end boundary, same-day events vanish | **1 failed** |

All four are regressions someone could plausibly introduce, and none of them could have been introduced quietly again. That is what we were missing.

Closes #66 — the issue that existed because on 2026-07-30 the agent said the calendar was clear while a full Thursday sat in Calendar.app.

One cosmetic note for next time, not worth a round-trip here: 14 lines carry trailing whitespace. `ruff check` passes (we do not enable W291), so nothing was blocking.

PR #102, opened 2026-08-12
Strip fabricated "[tools used: ...]" tags from text-only replies

mr-betterman:
## What

Live bug found 2026-08-12 in the dashboard chat_log: two turns on qwen-max
(openai wire format) claimed a tool ran and succeeded, but nothing was
actually written anywhere.

- `支付宝额基金賺了173` → reply said the +173 CNY was logged, with a fabricated
  `[tools used: log_pnl(...) -> Logged 支付宝基金 2026-08-12: +173 CNY
  (state.db, finance_entries)]` tag appended. No row for it in
  `finance_entries`.
- Correcting a calendar event's company name (EAPM → EPAM) → reply claimed
  `create_event` ran and synced to Apple Calendar. `state.db`,
  `calendar.ics`, and the Apple Calendar event all still said EAPM.

## Why

The `[tools used: ...]` tag is meant to be harness-only —
`session.add_exchange` appends it after a turn, built strictly from the real
`tool_calls` the loop executed. But the SOUL prompt tells the model "your
history shows `[tools used: ...]` lines for past turns", so the model has
seen the exact format in its own context. On both buggy turns the model
never emitted a `tool_use` block (`meta.tools` was empty both times) — it
just wrote the tag as plain text, imitating the harness instead of calling
the tool, and the fabrication rode straight through `session.add_exchange`
(which only appends its own tag when `tool_calls` is truthy) into
`chat_log` and the user's screen.

## Fix

In `run_loop`'s no-tool-uses branch, `tool_uses` is authoritative — if it's
empty, nothing ran this iteration, so any `[tools used: ...]` suffix in the
model's own text is a lie by construction. Strip it and emit a
`fabricated_tool_trace` observer event so tracing/dashboard can surface the
anomaly instead of silently eating it.

## Testing

`evals/deterministic/test_fabricated_tool_trace.py` adds two cases:
- a text-only reply carrying the fake tag (must be stripped, must not reach
  `chat_log`, must not fool the caller into thinking anything landed)
- a real `tool_use` call (must still get its honest tag from
  `add_exchange`, unaffected)

Full deterministic suite: 583 passed, 0 failed (53 pre-existing skips).
Judge evals didn't run in my sandbox (`deepeval` not installed, unrelated to
this change) but deterministic is the required gate per `CLAUDE.md`.

The two corrupted records this bug produced were repaired manually out of
band: the missing 支付宝基金 +173 CNY `finance_entries` row, and the
EAPM → EPAM rename in `calendar.ics` + the Apple `Waku` calendar event.


ShenSeanChen:
Thanks @mr-betterman — and sorry for the three days of silence.

**The guardrail is excellent and we want it.** I reproduced the bug on `main` with a stub model that emits text only, no `tool_use` block anywhere:

```
main:
  Logged it for you — 支付宝基金 是 +173 CNY 今天。
  [tools used: log_pnl(...) -> Logged 支付宝基金 2026-08-12: +173 CNY (state.db, finance_entries)]

your branch:
  Logged it for you — 支付宝基金 是 +173 CNY 今天。

both:  tools actually executed: []
```

Your diagnosis is exactly right, and the reasoning is the part I want to call out: `tool_uses` being empty is not "this looks suspicious", it is **a lie by construction** — the branch is only reached when nothing ran. That means the strip cannot misfire on a genuine tag, because a genuine tag never reaches this branch. Emitting `fabricated_tool_trace` instead of silently swallowing it is the right second half.

## The ask

The PR is 21 files. The fix is 2. I would like to merge the fix this week, and I cannot merge it in this shape — so could you reduce this PR to just the guardrail?

**Keep:**
- `waku/loop/agent.py`
- `evals/deterministic/test_fabricated_tool_trace.py`

**Please drop:**
- `waku/tools/finance.py`, `waku/tools/interviews.py`, and their registration in `waku/tools/__init__.py`
- the `finance_entries` / `interview_entries` tables in `waku/db.py`
- the Finance dashboard page — `waku/ops/static/{index.html,style.css,js/main.js,js/views.js}`, `waku/ops/dashboard.py`, `waku/ops/scoring.py`
- their tests, and `docs/superpowers/**`

## Why, honestly

Not because the tools are bad — they are not. The `overwrite` confirmation flow, and matching an interview process on company alone so a corrected role updates the row instead of forking it, are both more careful than most tool PRs we get.

It is that **every registered tool ships in every prompt, for every user**. `CONTRIBUTING.md` has a footprint ladder for exactly this, and a new core tool is its last rung. `log_pnl` would give every waku installation a tool whose account enum is `A股 / 支付宝基金 / 雪球基金 / IBKR / BIT`, and every `state.db` two tables for tracking one person's portfolio and job search. That is a config surface waku does not have yet, not a capability the core is missing.

Where they could live instead: your own fork (simplest, and they are already working code), or `examples/` as a self-contained "adding your own tools" walkthrough — that directory is teaching material and this is a genuinely good worked example of the pattern.

If splitting is more work than you want to spend, say so and I will land the two-file version with you as the author. **The fix is too good to hold hostage to a scope conversation** — I just cannot take the other nineteen files with it.

PR #144, opened 2026-08-18
ui: add a System / Light / Dark theme toggle to the dashboard

### LINKED ISSUE #138: Add a button for light/dark theme toggle
**What you're trying to do** — Add a toggle button to change the theme (dark/light).

The dashboard already has a light palette, but it only follows the OS setting.
There's no way to pick a theme in the page itself.

Ask: a small button that cycles System / Light / Dark and remembers the choice.
System stays the default, so nothing changes unless you click it.

**Where it belongs on the footprint ladder** — none of these. It's a button and
some CSS in `waku/ops/static/`. The model never sees it, so it costs nothing in
the prompt.

- [ ] a skill (`SKILL.md`, no Python)?
- [ ] a CLI + a README the model reads when it needs it?
- [ ] a tool behind an optional extra?
- [ ] a gateway (one file, text in and out)?
- [ ] something that genuinely has to live in the core?

**Who else needs this** — anyone running a dark desktop who wants to read the
dashboard in light, or the reverse. Also useful when recording.

**Would you want to build it** — yes

Aryan4443:
Closes #138.
 what this is

A small button in the page header that cycles **System → Light → Dark** and
remembers the choice in `localStorage`. System stays the default, so nothing
changes for anyone who never clicks it.

How it works

`js/theme.js` holds the *choice* (`system` / `light` / `dark`) and writes the
*resolved* theme to `data-theme="light|dark"` on `<html>`. Keeping those two
apart is the whole design.

Three details worth calling out, since each was a decision rather than a default:

- **It loads in `<head>`**, unlike every other script in `js/`. The attribute has
  to be on `<html>` before the first paint, or a Dark user gets a white flash on
  every reload.
- **`style.css` now has no `prefers-color-scheme` queries at all.** Since
  `theme.js` resolves System itself, the palette reads one switch instead of two
  sources that have to agree — and a media query left behind would silently
  outrank a manual pick, which reads as a half-dark page you cannot explain from
  either file alone. All five old blocks (palette, `.mdcode`, `.mdcode-head`,
  `.sqlbox`, the Connections tweaks) moved to `:root[data-theme="dark"]`.
- **`color-scheme` is explicit per theme**, no longer `light dark`. Otherwise
  native scrollbars and form controls keep following the OS — the one thing the
  button exists to override.

`localStorage` access is wrapped: it throws in private mode and with site data
blocked, and a theme button is not worth taking the dashboard down for. System
also keeps tracking the OS while the page is open, via a `matchMedia` listener.

All test passed , checked in local server buttons's working fine

### DIFF
--- a/evals/deterministic/test_theme_toggle.py
+++ b/evals/deterministic/test_theme_toggle.py
@@ -1,0 +1,68 @@
+"""DETERMINISTIC EVAL — the dashboard theme switch has exactly one switch.
+
+The dashboard used to follow the OS only, through `prefers-color-scheme`
+queries. The System / Light / Dark button works by resolving the choice in
+`js/theme.js` and stamping `data-theme="light|dark"` on `<html>`, so the CSS
+reads ONE attribute.
+
+That only holds while nobody adds a media query back. A stray
+`@media (prefers-color-scheme:dark)` would win over the picked theme for
+whatever it styles, and the symptom is a half-dark page nobody can explain from
+reading either file alone — so pin the invariant here instead.
+
+The other two checks cover the ways the button dies silently: the script has to
+load in <head> (or the page flashes the wrong palette on every reload) and the
+button has to be wired to a real handler."""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+STATIC = Path(__file__).resolve().parents[2] / "waku" / "ops" / "static"
+INDEX = (STATIC / "index.html").read_text()
+CSS = (STATIC / "style.css").read_text()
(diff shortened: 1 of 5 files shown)


ShenSeanChen:
Thank you @Aryan4443 — a System / Light / Dark toggle shipped in #192 as part of adopting the Waku Memory design system, which also moves the palette onto `data-theme` the same way this PR proposed. Closing so the two don't compete; your write-up of why the attribute has to land before first paint is exactly the reasoning that version follows.

PR #146, opened 2026-08-18
fix: let GPT-5.6 tool calls work on Chat Completions

emranemran:
## Summary
- Follow-on to #136: keep `gpt-5.5` as the OpenAI default (Completions + tools already work).
- GPT-5.6 sol/terra/luna default to `reasoning.effort=medium`, which **400s** on function tools on `/v1/chat/completions`. OpenAI's documented Completions escape hatch is `reasoning_effort=none` (otherwise: Responses API).
- The OpenAI adapter now sends `none` when the model id starts with `gpt-5.6` **and** tools are on the request. Gate/summarizer calls omit tools, so Luna can still reason.
- Lets `WAKU_MODEL=gpt-5.6-sol` work without rewriting the loop onto Responses.

## Test plan
- [x] `pytest -q evals/deterministic/test_models.py::test_gpt56_tool_calls_force_reasoning_effort_none` (offline)
- [x] Default path unchanged: `WAKU_PROVIDER=openai` with blank `WAKU_MODEL` still uses `gpt-5.5` and books an event
- [x] Opt-in: `WAKU_MODEL=gpt-5.6-sol` + `WAKU_SMALL_MODEL=gpt-5.6-luna`, then `Remember that Alex prefers morning meetings` → `save_note` succeeds (no 400)
- [x] Gate still prints `gate · skip/retrieve` on a 5.6-luna call (no tools, no forced `none`)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -177,6 +177,8 @@ WAKU_PROVIDER=anthropic
 # Each provider has sensible defaults (see waku/loop/models.py PROVIDERS):
 #   anthropic:  claude-sonnet-5              + claude-haiku-4-5 (gate/summarizer)
 #   openai:     gpt-5.5                      + gpt-4.1-mini
+#               (gpt-5.6-* need reasoning_effort=none for Completions+tools;
+#                set WAKU_MODEL=gpt-5.6-sol and the adapter does that)
 #   gemini:     gemini-3.5-flash             + gemini-3.1-flash-lite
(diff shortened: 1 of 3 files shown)


emranemran:
Hi @ShenSeanChen, I was trying to use the latest chatgpt reasoning models. Is this change alright as a follow-on to your fix from https://github.com/ShenSeanChen/waku-agent/pull/136?

PR #147, opened 2026-08-20
test: add deterministic WhatsApp webhook security tests

### LINKED ISSUE #101: Add deterministic WhatsApp webhook security tests
## What

The WhatsApp gateway verifies `X-Hub-Signature-256` before processing webhook
payloads, but there is no deterministic test covering that security boundary.

PR #26 specifically called for a regression test ensuring that a forged POST
gets a 403 and never reaches `waku.respond()`.

## Proposed scope

Add `evals/deterministic/test_whatsapp.py` with offline coverage for:

- missing or invalid signatures return 403 and never invoke the agent
- a correctly signed text message returns 200 and invokes the agent once
- an invalid webhook verification token returns 403
- `WHATSAPP_ALLOWED_PHONE` rejects non-allowlisted senders
- malformed and non-text payloads do not invoke the agent

All network calls and agent responses will be stubbed. No Meta credentials,
external requests, or new dependencies.

I’d like to take this. Please assign it to me before I start.

wuisabel-gif:
Closes #101.

## What

Adds `evals/deterministic/test_whatsapp.py` — offline coverage for the webhook security boundary that PR #26 called out but left untested.

## Coverage

| Case | Status | Agent invoked? |
|---|---|---|
| POST with missing `X-Hub-Signature-256` | 403 | never |
| POST with forged signature (wrong secret) — the PR #26 regression | 403 | never |
| Correctly signed text message from the allowed phone | 200 | exactly once, reply sent |
| GET verification with wrong `hub.verify_token` | 403 | — |
| GET verification with the configured token | 200, challenge echoed | — |
| Signed message from a non-allowlisted sender (`WHATSAPP_ALLOWED_PHONE`) | 200 (Meta still gets its ack) | never, nothing sent |
| Signed malformed JSON | 400 | never |
| Signed non-text (image) message | 200 | never |

## How

The handler is a stdlib `BaseHTTPRequestHandler`, so the tests serve it in a real `ThreadingHTTPServer` on `127.0.0.1:0` — loopback only, real HTTP round-trips. `Waku`, the DB connection, settings, and the Meta Cloud API send are stubbed via `monkeypatch`. No Meta credentials, no external requests, no new dependencies.

## Review notes from #101, addressed

- **Test at the `_build_handler` seam:** `WHATSAPP_ALLOWED_PHONE` is read by two entry points — `main()` (whatsapp.py:250, standalone) and `start_in_background()` (:329, dashboard) — and both build their handler through `_build_handler(...)` (:272, :330). The fixture drives `_build_handler` directly, so both entry points are covered at their shared seam; nothing tests only one module-level path. This is now also documented in the test module docstring.
- **Assert the agent was not invoked, not just the 403:** `_FakeWaku` stubs `respond()` and records every call. The forged/missing-signature tests assert both `== 403` and `calls == []` (call count zero, plus nothing sent) — the security boundary, not just the status code.
- **Mechanisms verified where expected:** signature check exercised at :187/:193, verify-token check at :163/:171, both through real HTTP round-trips.

One subtlety, documented in the module docstring: `do_POST` acks 200 *before* processing messages (Meta disables slow webhooks). Because the handler speaks HTTP/1.0 with no `Content-Length`, the response body is close-delimited, so a client that finished reading knows processing completed — the assertions are exact, not racy.

## Verified

- `pytest evals/deterministic/test_whatsapp.py` → 8 passed
- `pytest -q evals/deterministic` → 623 passed, 60 skipped (live-eval skips), no regressions
- `ruff check` clean


wuisabel-gif:
Replying to the review notes on #101 — both points are baked in:

**Two entry points, one seam:** confirmed `WHATSAPP_ALLOWED_PHONE` is read at :250 (`main`) and :329 (`start_in_background`), and that both funnel into `_build_handler(...)` (:272, :330). The fixture drives `_build_handler` directly, so the allowlist tests cover both the standalone and dashboard paths at their shared seam — no test touches only one module-level entry point. This is now spelled out in the test module docstring too.

**The assertion that matters:** the forged/missing-signature tests stub `respond()` with a call recorder and assert `calls == []` alongside the 403 — status code and agent-invocation are asserted as two separate things, exactly as you framed it.

PR #149, opened 2026-08-25
Add da-anomaly-analysis community skill

Terence699:
## Summary
- Add a community skill that walks through data-anomaly investigation (define the change, validate data, decompose contribution, then internal vs external causes).
- Frontmatter follows the Agent Skills format (`name` + `description` with trigger phrases) so the loader can match questions like "why did X drop".
- `python scripts/validate_skills.py` passes locally.

## Test plan
- [x] `python scripts/validate_skills.py` is green in CI (`skills-and-evals` succeeded)
- [x] Chat a matching prompt (e.g. "why did DAU drop") and confirm this skill body is loaded
- [x] Chat an unrelated prompt (weight change) and confirm it is not loaded


ShenSeanChen:
Thanks for this — the method itself is solid (the weighted-rate decomposition and the Simpson's paradox check are exactly right), and thank you for already dropping the `uv.lock` churn.

I pushed one small commit on top of yours (`32740e1`) that only changes the `description` line. Here's why.

The skill loader picks skills by word overlap: a skill loads when a message shares 2+ words of 3+ letters with its description. The old description included `"why did X drop"` and `"what caused this change"`, so `why`, `did`, `what` and `this` counted as matches, and the skill loaded on ordinary personal messages:

| Message | Loaded before |
|---|---|
| "why did Alex move our meeting" | da-anomaly-analysis |
| "my weight dropped this month, why" | da-anomaly-analysis |
| "why did my package get delayed" | da-anomaly-analysis |

I probed it with 4 messages that should trigger it and 10 everyday personal messages that shouldn't:

| | should fire | false positives |
|---|---|---|
| before | 4/4 | 7/10 |
| after | 4/4 | 0/10 |

The new description is anchored on metric vocabulary (DAU, GMV, conversion, retention, KPI, dropped, spiked…) instead of question words. I also checked that it doesn't take over any other skill's trigger phrases. The body is untouched, and the PR is still yours.

PR #153, opened 2026-08-27
Add SiliconFlow provider with China and Global endpoints

Terence699:
## Summary
- Add an OpenAI-compatible `siliconflow` provider (no new wire client). Default model is `deepseek-ai/DeepSeek-V4-Flash`.
- China (`https://api.siliconflow.cn/v1`) and Global (`https://api.siliconflow.com/v1`) are a `SILICONFLOW_BASE_URL` choice, same pattern as MiniMax/Kimi.
- Dashboard Connections, `.env.example`, pricing fallback, and a provider logo are included. Offline tests cover the openai-wire regional client and catalog URL.

## Test plan
- [ ] `python -m pytest -q evals/deterministic` (already 620 passed locally)
- [ ] `python scripts/generate_env_example.py --check`
- [ ] Set `WAKU_PROVIDER=siliconflow` + `SILICONFLOW_API_KEY`, chat once on the China default, then switch `SILICONFLOW_BASE_URL` to the Global URL
- [ ] Confirm Settings model picker lists `GET {base}/models` ids and a tool call still works

Live note: both catalog hosts returned HTTP 401 without a key (endpoints are up). This machine has no `SILICONFLOW_API_KEY`, so a real chat turn was not run.


Terence699:
Closing to reopen with an updated description: SiliconFlow China was live-tested after a local key was added. Same branch will come back as a new PR.

PR #154, opened 2026-08-27
Add SiliconFlow provider with China and Global endpoints

Terence699:
## Summary
- Add an OpenAI-compatible `siliconflow` provider (no new wire client). Default model is `deepseek-ai/DeepSeek-V4-Flash`.
- China (`https://api.siliconflow.cn/v1`) and Global (`https://api.siliconflow.com/v1`) are a `SILICONFLOW_BASE_URL` choice, same pattern as MiniMax/Kimi.
- Dashboard Connections, `.env.example`, pricing fallback, and a provider logo are included.

## Test plan
- [x] `python -m pytest -q evals/deterministic` — 620 passed locally
- [x] `python scripts/generate_env_example.py --check`
- [x] Live China default (`api.siliconflow.cn`): catalog lists 95 models including the default; a full Waku loop turn replied `env-ok`; a `save_note` tool call was emitted with the requested text
- [x] Live Global (`api.siliconflow.com`) with the same China-issued key: catalog and chat both return 401 `Token is invalid` — expected for a regional key; switch `SILICONFLOW_BASE_URL` after using a Global console key
- [ ] Settings model picker lists `GET {base}/models` ids in the dashboard


Terence699:
Withdrawing to rebase onto the latest upstream main and finish the Global-endpoint live test before opening again.

PR #157, opened 2026-08-27
Add Firecrawl as a search_web backend

Terence699:
## Summary
- Plug Firecrawl Search into the existing `search_web` tool (stdlib `urllib`, `POST https://api.firecrawl.dev/v2/search`). No new tool name, so it does not occupy extra prompt space.
- Connections gets a Firecrawl card with only `FIRECRAWL_API_KEY` — same shape as Tavily. If that key is set, `search_web` uses Firecrawl; otherwise Tavily; otherwise DuckDuckGo.
- Offline tests pin backend selection and the Firecrawl request shape.

## Test plan
- [x] `python -m pytest -q evals/deterministic/test_search.py evals/deterministic/test_integrations.py evals/deterministic/test_static_assets.py`
- [x] `python scripts/generate_env_example.py --check`
- [x] Live `search_web` with `FIRECRAWL_API_KEY` returns `via Firecrawl` and real URLs
- [x] Firecrawl key present wins over Tavily (offline); Connections card has no backend dropdown
- [ ] Restart dashboard, confirm Firecrawl is key-only, then chat-search once


Terence699:
Closing: having two connected search cards with a hidden priority (Firecrawl over Tavily) is misleading. search_web should not imply both engines are in use when only one runs.

PR #179, opened 2026-09-07
Dashboard: changed the chatbar's input tag to textarea so that it can expand vertically 

Twishha-Soni:
**What this does, and why**
The issue was on dashboard, when i was writing my prompt, I need to do horizontal scroll to check what I typed, so I changed it to vertical expansion instead. I also observed when hit shift + enter, instead of going to new line, it was by default sending the query to the agent, so changed that as well.

**Which issue does it close?** (`Closes #123`)
The issue was with .chatbar html class with the input tag which cannot expand horizontal, hence changed it to textarea tag.
---

- tested it with manually  messaging the agent running on localhost.
- [ ] **A deterministic eval** in `evals/deterministic/` covers the behavior.
- [ ] `make gate` and `make lint` pass locally.
- [ ] Any heavy or optional dependency is **behind an extra**, not in the
      default install.
- [ ] No hidden network calls, no reading secrets or `.env`, nothing runs at
      install time.

- I tested it manually on the browser, I haven't done any eval or testing as it was the change in the frontend mostly.
```
```

**Anything you're unsure about?** Say so. A question here is cheaper than a
review round.
- I am not sure about wether the addition in the javascript sendchat function I made of retaining the message when something went wrong.
---
First PR here? CI needs a maintainer to approve your workflow run. If it looks
stuck, say so on the PR — that delay is ours, not yours.


ShenSeanChen:
Thank you for this, and I'm sorry it sat for as long as it did. You found a real bug, you found it first, and you diagnosed it correctly: an `<input>` cannot expand, and the fix is a `<textarea>`.

You also spotted the Shift+Enter behaviour on your own, which is the part most people miss. It is not obvious until you try to write a second line and the message sends instead.

I have merged #210, which fixes the same thing. I want to be straight with you about why, because "someone else got there" is not the reason.

#210 arrived two weeks after yours, and what it carries is the evidence: nine deterministic evals covering the swap. That matters here because `AGENTS.md` makes a deterministic eval a hard requirement for a behaviour change, and I test a claimed regression test by reverting each source file and checking the test actually fails. Yours was correct code without that scaffolding; the other was correct code with it. Asking you to add it now would have meant asking you to rebuild something that already existed in another branch, which felt worse than telling you plainly.

Two things I would rather you take from this than a closed PR.

The diagnosis was right, and being first counts. I have said so in the merge context so it is on the record, not just in this thread.

And the gap was process, not skill: an eval, and the checklist boxes for `make gate` and `make lint`. Those are learnable in one PR. If you want another, the `good first issue` list is open, and I am happy to point you at one that fits what you already showed you can do here. Say the word and I will pick one out.

Sorry again for the wait. It was a good catch.

PR #185, opened 2026-09-10
fix(dashboard): write SOUL.md as UTF-8 on save_soul

### LINKED ISSUE #173: Dashboard save_soul should write SOUL.md as UTF-8 on Windows
This is a follow-up to [Issue #23](https://github.com/ShenSeanChen/waku-agent/issues/23) and the related fixes in [PR #25](https://github.com/ShenSeanChen/waku-agent/pull/25) and [PR #28](https://github.com/ShenSeanChen/waku-agent/pull/28).

Issue #23 reported Windows text-encoding problems. PR #25 added explicit UTF-8 encoding to dashboard reads, and PR #28 covered text file I/O outside the dashboard. The dashboard `save_soul` path still writes `SOUL.md` without an explicit encoding:

```python
(settings.home / "SOUL.md").write_text(text + "\n")
```

On Windows, this uses the platform default encoding. Saving non-ASCII content, such as Chinese text or emoji, can fail or leave `SOUL.md` in an encoding that the runtime cannot read consistently. The runtime already reads this file as UTF-8.

Expected behavior:

- Dashboard saves should write `SOUL.md` as UTF-8.
- Non-ASCII content should be preserved when the file is read by the runtime.

The fix is to pass `encoding="utf-8"` to this `write_text()` call and add a deterministic regression test. This is a separate write-path gap from the read and non-dashboard paths covered by the earlier fixes.

ZxlDragonDoctor:
### Problem
Dashboard `save_soul` wrote `SOUL.md` with `Path.write_text(...)` and **no explicit encoding**. On Windows the platform default codec can mangle non-ASCII content (Chinese, emoji), while the runtime and `save_skill` already read/write UTF-8.

### Solution
Pass `encoding="utf-8"` to the `save_soul` write so it matches `save_skill` and the runtime reader. Add a deterministic regression test that fails if any `write_text` call omits encoding.

### Issue alignment
- No maintainer design note on #173; implements the exact fix requested in the issue (`encoding="utf-8"` + regression test).

### Testing
- Command: `python -m pytest evals/deterministic/test_skill_encoding.py -q`
- Result: `5 passed`
- New case: `test_dashboard_save_soul_is_written_as_utf8` — non-ASCII body round-trips as UTF-8 under the repo's `require_explicit_text_encoding` fixture.

### Agent dimension
context-management — durable persona file (SOUL.md) must survive Windows encoding defaults.

### Core value
Windows users editing the agent persona via the dashboard could corrupt SOUL.md silently; one-line encoding pin closes it.

Fixes #173


ZxlDragonDoctor:
Pushed `aa095d3` to resolve the merge conflict with main. Kept the UTF-8 SOUL coverage from this PR and merged main's renamed `test_dashboard_soul_is_written_as_utf8` + exact-content assertion into one test that still requires explicit UTF-8 writes and covers Chinese + emoji body.

ZxlDragonDoctor:
Resolved the merge conflict with current `main` (aa095d3). The upstream test was renamed to `test_dashboard_soul_is_written_as_utf8`; this branch keeps that name and still asserts CJK + emoji UTF-8 round-trip via explicit `encoding="utf-8"` fixture.

ShenSeanChen:
Thanks @ZxlDragonDoctor — you found a real bug. The same fix landed in #174 (`save_soul` now writes `SOUL.md` as UTF-8, with a regression test that fails on the old line), so what's left here is a variation of that test. Closing as superseded, and thank you for tracing it: a Windows encoding bug on that path is exactly the kind of report we want.

PR #189, opened 2026-09-12
dashboard: adopt the Waku Memory design system (1/3: foundation)

etsai0603:
Part 1 of 3. Spec: `docs/dashboard-design-system-design.md`.

## What this is

The dashboard now uses the Waku Memory design system: bone and charcoal, one amber accent, Instrument Sans / JetBrains Mono / Playfair Display SC, 2px corners, no shadows, and buttons that never fill. Layout does not move — every element is where it was.

- `waku/ops/static/design/` holds Memory's `tokens.css` and `controls.css`, copied unchanged by `scripts/sync_design.py`, which records their source commit and sha256 in `SOURCE.md`.
- The three fonts are served from the package (`static/fonts/`, OFL licenses included). Nothing is fetched at runtime. `dashboard.py` gains one line so `.woff2` is served as `font/woff2`.
- `style.css` no longer holds a raw colour, size, weight, corner, shadow or duration. Its old short names (`--ink2`, `--line`, …) are aliases to tokens for now; PR 3 removes them.
- `util.js` stamps `data-slot` on native controls so `controls.css` gives every button and field the same focus ring, pressed and disabled states.

## Why

The dashboard and the Memory console are one product family but looked like two products. It should also stay on the system as people add screens: `static/README.md` gets a "Design system" section, `CLAUDE.md` points to it, and `evals/deterministic/test_design_system.py` (20 checks) fails on a hand edit to a copied file, any raw colour/size/weight/corner/shadow/duration in `style.css` or a `js/` inline style, amber used as text, a filled button, and any new file using the old variable names.

One bug found while checking: Playfair Display SC draws `fi`/`fl` ligatures as lowercase glyphs ("WORKfLOWS"). Titles now turn ligatures off, with a regression test.

## How to test

```bash
pytest -q evals/deterministic          # 697 passed
make dashboard                          # then open every tab, light and dark
```

Checked here: all 13 tabs in both themes, zero console errors, fonts load as `font/woff2`, and the wheel contains `static/design/` and `static/fonts/`.

## Please decide

- **License.** `design/tokens.css` and `design/controls.css` come from the private `waku-memory-frontend` repo. The same files are already public at https://www.waku.one/design/, but in this repo they would fall under MIT. If that is not OK, I'll add a note to `design/SOURCE.md` that `design/` is excluded from MIT.
- **Overlapping open PRs.** #144 adds a theme toggle, which PR 2 of this series also does — worth picking one. #179 and #102 edit `style.css` and #178 edits `models.js`; whichever lands second needs a small rebase, and the new test will ask them to use tokens.

## Next

- PR 2 — shell: sidebar as Memory's nav rail, page header, chat dock, light/dark toggle.
- PR 3 — views: each tab on the closest Memory pattern, shared primitives in `js/ui.js`, spacing on `--space-*`, old names removed.


ShenSeanChen:

PR #190, opened 2026-09-12
dashboard: adopt the Waku Memory design system (2/3: shell)

etsai0603:
Part 2 of 3. **Stacked on #189 — please review #189 first.** The first 11 commits here are #189; this PR adds the last 4 (from `docs: design system spec — PR 2 shell in full…`). Once #189 merges, only those 4 remain. Spec: `docs/dashboard-design-system-design.md` §11.

## What this is

The dashboard's frame — sidebar, page header and chat dock — is rebuilt after the Waku Memory console's shell (`components/console/console-nav.tsx`), and gains a theme toggle.

- **Sidebar → nav rail.** 196px on a paper ground, mono uppercase items, a 2px amber rule on the current page, faint counts that show nothing at zero. The three group labels stay as faint captions (13 items, where Memory has 5). It collapses to 56px and each item's first letter; the full name stays in `aria-label` and the tooltip. The collapse is not remembered, so a returning reader always gets the navigation back.
- **Removed:** drag-to-resize on the sidebar, the hide/reopen buttons and their `navW`/`navHidden` storage, and the `provider · model` line (the chat dock's model chip already shows and switches the model).
- **Theme toggle** in the rail's bottom row: one button cycles system → light → dark, stored under `waku-theme` (the key the Memory console uses) and set as `data-theme` on `<html>`. A short inline script in `<head>` applies it before the design files load, so the wrong theme never flashes.
- **Chat dock** takes the paper ground, like the rail, and keeps its resize handle and collapse button. The page title sits in a row with room for actions (filled in PR 3).
- **Below 768px** the rail becomes a row across the top, and the dock opens over the page.

## Why

These are the design owner's decisions from a side-by-side mockup of every open question for PRs 2 and 3; spec §10 and §11 record them.

## How to test

```bash
pytest -q evals/deterministic          # 702 passed
make dashboard                          # click the theme icon (bottom-left) and the ‹ collapse button; narrow the window below 768px
```

Checked here: all 13 tabs in light and dark, zero console errors, `theme.js` in the wheel. New checks in `test_design_system.py`: the theme is applied in `<head>` before the design files, the toggle is wired, every rail link carries its letter and full name, and nothing of the old resizable sidebar remains.

## Please note

- **#144** also adds a theme toggle. This one uses Memory's three states and storage key; worth keeping only one.

## Next

PR 3 — views: stat band, line tabs, Notice boxes, one-hue charts, shared primitives in `js/ui.js`, spacing on `--space-*`, old variable names removed.


ShenSeanChen:

PR #191, opened 2026-09-13
dashboard: adopt the Waku Memory design system (3/3: views)

etsai0603:
Part 3 of 3. **Stacked on #190 (which is stacked on #189) — please review those first.** This PR adds the last 4 commits (from `docs: design system spec — PR 3 views and primitives`). Spec: `docs/dashboard-design-system-design.md` §12.

## What this is

Every tab is rebuilt from Waku Memory's primitives, and spacing moves onto Memory's steps.

- **`js/ui.js`** — Card, Badge, Table, Tabs, Notice, a stat band, Row and a native `<dialog>`, each a function returning HTML the way the views already build markup. Names follow Memory's `components/ui` with a `ui` prefix (the `js/` files share one global scope). The static README lists which to use when.
- **Views** — stat tiles become the stat band; sub-tabs the line tabs (ink bar; amber stays in the rail); explanation boxes Notices; every table one Table; fifteen kinds of chip, pill and status tag one Badge with Memory's variants (`ok`, `warn`, `bad`, `live`, `miss`, `value`, `neutral`). The retrieval gate shows two figures over a thin bar; the model-race scatter uses one hue with each grade printed. The provider and connection dialogs are native `<dialog>`s (Escape and the scrim close them).
- **Spacing** — `--space-2/3/4/6/8` between things, `calc(var(--spacing) * 0.5|1|1.5)` inside one control, so changing `--spacing` moves the whole dashboard. The short aliases from #189 (`--ink2`, `--line`, …) are replaced and deleted. CSS nothing emits any more is deleted.

## Why

These are the design owner's decisions for PR 3 (spec §10). The new checks keep later screens on the system: `test_design_system.py` fails on raw-px spacing, any old variable name, and hand-written Badge/Table/Tabs/Notice/stat-band/Row markup.

## How to test

```bash
pytest -q evals/deterministic          # 705 passed
make dashboard                          # every tab, light and dark; open Models → Edit and Connections → Configure
```

Checked here: all 13 tabs in light and dark, zero console errors, `ui.js` in the wheel.

## Small visible changes to know about

- The memory race's "told / then asked" table is now two tables back to back (the Table primitive has no in-table section row).
- Table headers are uppercase mono, as in Memory — including the store names in the memory race.
- Arena card stat badges no longer highlight the active sort; the sort buttons still do.

## Not in this PR

- Unused JS found along the way, left untouched: `yourModelsCard`, `loadModelList`, `renderCatalog`, `renderCatalogList`, `modelRow`, `pickTrack`.
- Cards (47 still hand-written), buttons, the `<a class="reveal">` actions, the dropdown menus and line heights are the next PR.


ShenSeanChen:

PR #192, opened 2026-09-13
dashboard: cards, buttons, actions and menus from ui.js

etsai0603:
Stacked on #191 — review that first; this PR's own changes are the last three commits.

## What

PR 3 moved every value onto tokens but left some markup hand-written. This PR moves the rest onto the primitives in `js/ui.js`, so a change to a primitive reaches every screen:

- **Actions are buttons (13A).** The 45 `<a class="reveal">` actions (edit, delete, grade, clear, open in Finder) are now tertiary `<button>`s, so the keyboard can reach them. Delete is the same button with `--bad` text. A link that goes to another tab stays a link (`uiLink`).
- **`uiButton`.** Memory's four levels (primary, secondary, tertiary, destructive) and two sizes. It replaces the 27 buttons that had their own classes (`save`, `save ghost`, `sessbtn`, `msg-copy`, …).
- **Cards (14A).** The 47 hand-written cards now call `uiCard`. When a card's title is a name (a provider, a store, a memory pillar), it uses the display face, as Memory's Card does.
- **Menus (15B).** The chat history, the model menu and the model picker now share `openMenu`: a paper ground with a hard edge, a raised surface under the hovered item, and amber text on the current one. Escape and an outside click close it, the arrow keys move between items, and focus goes back to the trigger.
- **Line height (16B).** `--leading-normal` becomes 1.55 (Memory has 1.72). The other raw line heights move onto `--leading-*`.
- **Headline figures** in the stat band and the gate are `--text-lg`, as a follow-up to PR 3. That change is on #191 too.

The old button and menu rules are deleted (8 unused classes, 14 rules). `test_design_system.py` now fails if the page has a hand-written card, a hand-classed button, an `<a class="reveal">`, or a raw line height.

## Waku Memory follows

The Memory stat band at 20px, the 15B menu look and line height 1.55 are going to Waku Memory in separate PRs against `spec-driven`. Until they merge, `type.css` and `design/SOURCE.md` note that `--leading-normal` is ahead of Memory's `globals.css`.

## Notes

- `.cmp-col` (arena columns) and `.provcard` (provider cards) change layout slightly because they now sit inside `uiCard`'s body.
- Unused JS left alone, as in #191: `yourModelsCard`, `loadModelList`, `renderCatalog`, `renderCatalogList`, `modelRow`, `pickTrack`.

## Verified

- `pytest evals/deterministic`: 707 passed, 61 skipped. `ruff check .` is clean, and `node --check` passes on every file in `js/`.
- All 13 tabs render in light and dark with zero console errors (headless Chrome; sheets attached).
- Menus and dialogs need a click, so they were checked by hand on port 7778.

## Screenshots


ShenSeanChen:
Three small commits on top of yours, from reviewing this with Sean on localhost:

1. **The architecture chart's border clears its top label.** "reply, out the same gateway" sat on the dashes; the border moves up 10 units and the title 6.
2. **Waku's chat answers keep their card.** Sean asked for the answer box back. It's square and on the page ground, so the two sides still read apart: the rounded bubble is you, the box is Waku.
3. **The chat dock sits beside the page and yields width to it.** With the dock dragged wide, main dropped below the chart's 760px floor and the chart's right third scrolled out of view (47% at a 1280px window). The dock now keeps its dragged width while main has `--main-min` (480px of chart plus padding) and shrinks toward 260px when it doesn't; the drag handle stops at the same limit, and the chart scales down to 480px instead of scrolling.

Measured with the dock open at 885 (rail collapsed) to 1920px: the dock never covers main, and the chart is whole at every width. The one case left: about 885px with the rail open, where rail + 260px dock + main's minimum don't fit, and 120px of the chart scrolls. Auto-collapsing the rail below that width would close it; that's your call, since the rail's collapse is deliberately not remembered.

`test_design_system.py` and `test_static_assets.py` pass (36).

ShenSeanChen:
One more, from the same session: **chat content stays inside a narrow dock.** At the dock's 260px floor the badge row didn't wrap, `tool · waku_memory_memory_search` (318px) was wider than the whole 254px column, and a long word in a message spilled past its bubble, so the column scrolled sideways (content 898px in a 254px window at first).

- `.stages` wraps; in that row a badge is capped at the column and cut with an ellipsis.
- `uiBadge` wraps its text in `.badge-t` — text sitting directly in an inline-flex badge is an anonymous flex item that `text-overflow` cannot reach. Nothing else changes: across #models, #ops and #loop (68 badges) every badge's text is intact, none is cut, and the cap is not applied outside the chat.
- `toolChip` passes the full tool name as the title.
- `.bubble` and `.r` get `overflow-wrap:anywhere`.

Measured with the real chat history plus a 250-character unbroken word in a bubble and a 300-character URL in a reply: no sideways scroll at 1024, 885 (rail collapsed) or 1440. Design and static tests pass (36).

PR #193, opened 2026-09-13
feat: add local OpenCode provider with Muse support

huseyiner6:
**What this does, and why**

Adds an opt-in `opencode_local` provider for people who already use OpenCode. Waku connects to its local HTTP server, keeps its own memory and tool loop, and leaves provider credentials in OpenCode. The default model is Muse Spark 1.3 Contributor Free; explicit model choices are preserved and there is no automatic model fallback.

The adapter creates an owned session, disables OpenCode native tools, submits each attempt once through `prompt_async`, and polls for completion. It supports a 1800-second response deadline (configurable, including 0), bounded protocol repairs, usage aggregation, and session cleanup. Free-tier denials retain provider evidence and prevent the same client from resubmitting an unavailable model through gate/graph fallback paths.

The existing Models page supports keyless server configuration and lists connected models from `/provider`. Pricing marks the exact Muse contributor id as free and otherwise uses catalog rates or estimates. Setup and limitations are documented in `docs/opencode-local.md`. The transport uses the Python standard library; no dependency or new top-level package is added.

**Which issue does it close?**

None.

---

- [x] **Tested it, not just written it.** Offline transport, loop, quota, configuration, catalog, and model-picker checks.
- [x] **A deterministic eval** in `evals/deterministic/` covers the behavior and failure cases.
- [x] `make gate` and `make lint` pass locally.
- [x] No new heavy, optional, or core dependency.
- [x] No hidden network calls or secret discovery; nothing runs at install time. Server access occurs only for the selected provider or an explicit catalog/connection request. Configuration uses Waku's existing settings path.

```sh
make gate
make lint
PYTHONPATH=. python scripts/generate_env_example.py --check
```

- Gate: **833 passed, 61 skipped**, gate open. Judge evals were skipped because no provider API key was supplied.
- Lint: all checks passed.
- Generated environment example: current.
- Checks used an isolated Waku home and no model credentials. Python 3.14 was selected through Make's `PY` override.

Coverage includes synthetic waits beyond 120 seconds without duplicate submission, owned-session quota failures, tool protocol repairs and usage, native-tool rejection, interruption during polling/cleanup, HTTP deadlines, real Waku loop execution with fake transport, and graph/retrieval fallback after quota denial. All test prompts, responses, and identifiers are synthetic.

**Anything you're unsure about?**

A live model smoke test has not been run for this contribution; provider behavior is tested with synthetic responses. OpenCode controls model access and free-tier availability. The adapter currently expects an unauthenticated loopback server, returns completed text rather than streaming, and cannot forward `max_tokens` because the prompt API does not expose that field.


huseyiner6 (commented):
.

PR #206, opened 2026-09-17
loop harness aware

khosimorafo:
The loop knows which project it is bound to. With `WAKU_HARNESS` set, the system prompt carries the project's name, its authority rung, and the map index with authority labels (index only, no bodies). Unbound Waku is unchanged.

`knowledge.py` moves from `ops/` to `runtime/` so the runtime does not import from ops.

Tests: 796 deterministic pass (4 new), lint clean.


khosimorafo:
Opened against the wrong repository by mistake (a fork-specific change; the fork had upstream added as a remote). Closing; apologies for the noise.

Issue #209, opened 2026-09-21
Dashboard chat input is a single-line <input> — long messages scroll out of view instead of wrapping


sathudeva7:
**What happened, and what you expected instead**

The chat composer in the dashboard (the "Message Waku…" box) is a single-line
control. When I type or paste a message longer than the visible width, the text
scrolls horizontally and I can only see the last fragment — I can't read back
what I've written before sending.

I expected it to behave like every other chat UI (ChatGPT, Claude, Gemini): the
box wraps the text and grows downward to a few lines, then scrolls vertically
once it hits a max height.

<img width="562" height="713" alt="Image" src="https://github.com/user-attachments/assets/33252e20-9603-4b77-8ba0-bf17cda5cfa3" />

Root cause looks like `waku/ops/static/index.html:66`:

    <input id="dmsg" placeholder="Message Waku&hellip;" autocomplete="off">

An HTML `<input>` can't wrap or grow — it's a one-line control by definition.
A `<textarea>` is needed.

There's already precedent in the repo: the Compare view's composer uses a
textarea (`waku/ops/static/js/compare.js:440`), so the dock chat is the odd one
out.

**How to reproduce** — the smallest sequence that shows it.

1. `make dashboard`, open http://localhost:7777
2. Click into the chat box on the right
3. Paste a 2–3 sentence message, e.g.
   "Imagine an e-commerce checkout flow. Without the Strategy pattern, adding
   new payment methods forces you to modify the Checkout class continuously,
   violating the Open/Closed Principle."
4. The box stays one line tall; the beginning of the message is clipped off the
   left edge and unreadable without arrowing back through it.

**Your setup**
- waku version / commit: 0.1.8 / 761c420
- Python version (`python -V`): Python 3.11.13
- OS: macOS 15.4.1 (arm64)
- Provider and model: `openai` / `gpt-5.4`
- Browser: Brave (Chromium 153)

**Trace or output** — not applicable, this is frontend-only. No backend request
is involved; it reproduces before the message is ever sent.

**Anything you already ruled out?**
- Not a CSS width issue — `.chatbar input` at `style.css:420` already sets
  `flex:1; min-width:0`, so the box fills the dock correctly. The element type
  is the limit, not its width.
- Not browser-specific — an `<input>` behaves this way everywhere.
- Not related to the dock being narrow; widening the dock just delays the clip.

**Notes for whoever picks this up**

[...]

PR #210, opened 2026-09-21
Make the chat composer wrap and grow instead of clipping long messages

sathudeva7:
**What this does, and why**

The dock's composer was `<input id="dmsg">`. An `<input>` is a single line by
definition, so anything longer than the box scrolled out of sight to the left:
you could not read back what you had typed before sending it. Found in live
use while dictating a multi-sentence message into the dashboard.

It is now a `<textarea rows="1">` with an `autogrow()` in `render.js`, capped
at five lines by a max-height and scrolling past that. Enter still sends;
Shift+Enter now inserts a newline, which the old `<input>` could not do.

Three things the swap forced, each a real defect seen in the browser rather
than guessed at:

- The styling rule selected `.chatbar input` only, so the textarea lost its
  border, padding and colour. Widened to match both.
- A textarea does not inherit the page font; without an explicit `--face-sans`
  the composer rendered in the browser's monospace.
- `.chatbar` had no `align-items`, so its default `stretch` was silently
  equalising the composer and both buttons. Once the composer could grow,
  stretch made the buttons grow with it. Pinning the row to `flex-end` then
  exposed that `#mic` carries no vertical padding while `.btn` gives Send 10px
  — 34px vs 17px, the mic visibly half-height. The buttons are now grouped in
  `.chatbar-actions` so they stretch to EACH OTHER, and the composer and that
  group floor at the same derived `min-height` so an empty bar is flush on load.

`stopMic()` calls `autogrow` too. A dictated sentence is written to `.value`
from code and a textarea never resizes itself, so without it voice reproduced
the identical bug through the mic.

**Which issue does it close?**  [#209]

---

- [x] I read `AGENTS.md`, and this change keeps its hard rules.
- [x] **A deterministic eval** in `evals/deterministic/` covers the behavior.
- [x] `make lint` passes. `make eval` (the deterministic half of the gate)
      passes. **The judge half of `make gate` was not run** — see "unsure
      about" below.
- [x] Any doc sentence this change makes false is fixed in this PR too.
      (Checked `static/README.md` and `docs/context/design-system.md`; neither
      describes the composer element, so nothing went stale.)

The 61 skips are pre-existing: live evals gated behind `WAKU_RUN_LIVE_EVALS=1`
plus an API key.

In the browser (`make dashboard`, hard-reload, console clean, zero errors),
three rounds — each one found the next defect:

1. Pasted a 3-sentence message. Before: clipped, only the tail visible. After:
   wraps to 4 lines, grows, stops at 5 and scrolls.
2. Caught the mic button at half the height of Send once the row stopped
   stretching. Fixed by grouping, re-checked.
3. Caught the empty composer sitting ~4px taller than both buttons on load.
   Fixed by the shared `min-height`, re-checked.

Also exercised: Shift+Enter inserts a newline, plain Enter sends and the box
shrinks back to one row, and a voice transcription through the mic button
lands in a correctly-sized box.

New eval: `evals/deterministic/test_static_composer.py`, 9 checks. There is no
JS test runner here by design, so these are text-level checks in the spirit of
`test_static_assets.py`. The first is the case that would have caught this
bug — `#dmsg` is not an `<input>`. The rest pin each regression above,
including that the two `min-height` expressions stay character-identical.


ShenSeanChen:
Thanks for this, and thanks for writing the issue first with the root cause and the three follow-ups already worked out. That made the review fast.

I tested it rebased onto current `main` and it holds up:

- 959 deterministic evals pass, `node --check` clean on both JS files, design-system evals 30/30 (every value you added is a token, no colour literals).
- I reverted each of your four changed files independently. Each one fails at least one of your tests (index.html 3, render.js 2, style.css 5, main.js 1). The eval bites in all four directions, which is more than most PRs here manage.

Three things I want to call out, because they are the parts that would have been missed:

- the textarea not inheriting the page font, so the composer would have silently rendered in monospace
- `.chatbar` having no `align-items`, and the mic/Send height mismatch that only showed up once the row was pinned
- `stopMic()` calling `autogrow`, since a dictated sentence is written to `.value` from code and reproduces the whole bug through the mic

Those read like things you found by using it rather than by reasoning about it, which is the right way round.

One thing before merge: the branch is based on `761c420` and is seven commits behind. `main` has since rewritten `sendChat` (a paused dashboard now has to resume when a message is sent), so I would rather the rebase happen on your side than have me squash over it.

```bash
git fetch origin && git rebase origin/main && git push --force-with-lease
```

It rebased cleanly when I tried it locally, so this should be a formality. Ping me when it is up and I will take another look and get it in.

